PR #12, opened 2019-03-12
pdfmorph function

chiahaoliu:
Make a python function which replicates the functionality in the original pdfmorph app.


CJ-Wright:
Thank you!!!

CJ-Wright:
@chiahaoliu is this ready to go in?

chiahaoliu:
@CJ-Wright not yet, I am inspecting an edge case I just found

chiahaoliu:
@CJ-Wright Just pushed my fix

CJ-Wright:
Missing matplotlib in requirements

CJ-Wright:
LGTM @pavoljuhas thoughts?

diffpy/pdfmorph/pdfmorph_api.py
```diff
@@ -0,0 +1,200 @@
...
+        else:
+            chain.append(morph_cls())
+        refpars.append(k)
+    # exclude fixed options
+    if fixed_operations:
+        if not isinstance(fixed_operations, Iterable):
+            fixed_operations = [fixed_operations]
+        for opt in fixed_operations:
+            refpars.remove(opt)
+        print("== INFO: Following morphing steps are fixed ==:\n{}\n"
```
pavoljuhas: please put this after `if verbose` or so.  It is better if function does not produce debugging output unless requested.

diffpy/pdfmorph/pdfmorph_api.py
```diff
@@ -0,0 +1,200 @@
...
+        minimizing targets during morphing. Default to False.
+    fixed_operations : list, optional
+        A list of string specifying operations will be keep fixed during
+        morphing. Default is None.
+    refine : bool, optional
+        Option to execute the minimization step in morphing. If False,
+        the morphing will be applied with parameter values specified in
+        `morph_config`. Default to True.
+    verbose : bool, optional
+        Option to print full result after morph. Default to True.
```
pavoljuhas: can we change the default to `verbose=False`?

diffpy/pdfmorph/pdfmorph_api.py
```diff
@@ -0,0 +1,200 @@
...
+                      baselineslope=None, qdamp=None)
+
+# map of operation dict
+# TODO: include morphing on psize
+morph_step_dict = dict(scale=morphs.MorphScale,
+                       stretch=morphs.MorphStretch,
+                       smear=[morphs.MorphXtalPDFtoRDF,
+                              morphs.MorphSmear,
+                              morphs.MorphXtalRDFtoPDF],
+                       qdamp=morphs.MorphResolutionDamping)
```
pavoljuhas: Are `default_config` and `morph_step_dict` constant?
If so maybe they should be in upper case or prefixed with `_` to mark them as module-internal variables.

diffpy/pdfmorph/pdfmorph_api.py
```diff
@@ -0,0 +1,200 @@
...
+        normal rw value
+    pcc : float
+        pearson correlation coefficient
+
+    Examples
+    --------
+    # morphing (xobj, yobj) pair to (xref, yref) pair with scaling
+    from diffpy.pdfmorph import pdfmorph, default_config, plot_morph
+
+    morph_cfg = dict(default_config)
```
pavoljuhas: can this be replaced with a function `default_config()`?
otherwise this risks users would change `default_config` module variable with unpredictable results.

diffpy/pdfmorph/pdfmorph_api.py
```diff
@@ -0,0 +1,200 @@
...
+        output = "\n".join(["# {} = {:.6f}".format(k, v) for k, v in \
+                rv_cfg.items() if v is not None])
+        output += "\n# Rw = {:.6f}".format(rw)
+        output += "\n# Pearson = {:.6f}".format(pcc)
+        print(output)
+
+    return chain, rv_cfg, rw, pcc
+
+
+def plot_morph(chain, fig=None, **kwargs):
```
pavoljuhas: It seems that this function updates only the `ax` axis object.
A better interface would be to have `plot_morph(chain, ax=None)` so that
user can plot to any specific Axis, which is however quite clunky now.
Creating `ax` with a `subplots(**kwargs)` call is not a business for `plot_morph` to do,
I would suggest to use `ax = pyplot.gca() if ax is None else ax`.

I would also return a list of created Line2D objects, which could be then conveniently formatted.
User already has the `fig` object or can get it as `ax.figure`.

diffpy/pdfmorph/pdfmorph_api.py
```diff
@@ -0,0 +1,200 @@
...
+    -------
+    chain : diffpy.pdfmorph.morphs.morphchain.MorphChain
+        instance of morph chain. ``xobj, yobj, xref, yref = chain.xyallout``
+        will return morphed results
+    rv_cfg : dict
+        dictionary of refined parameters
+    rw : float
+        normal rw value
+    pcc : float
+        pearson correlation coefficient
```
pavoljuhas: if there are so many returned objects, it is maybe better to return a dictionary or namedtuple.
I would have tough time to remember if the 3rd returned value is rw or pcc.

diffpy/pdfmorph/pdfmorph_api.py
sbillinge: yes, default verbose should always be false.

diffpy/pdfmorph/pdfmorph_api.py
sbillinge: 👍

CJ-Wright: I have a preference for `dict` in this case.

chiahaoliu:
@pavoljuhas Thanks for your feedback! I've pushed changes regarding comments

diffpy/pdfmorph/pdfmorph_api.py
```diff
@@ -20,17 +20,29 @@
...
-                       qdamp=morphs.MorphResolutionDamping)
+_morph_step_dict = dict(scale=morphs.MorphScale,
+                        stretch=morphs.MorphStretch,
+                        smear=[morphs.MorphXtalPDFtoRDF,
+                               morphs.MorphSmear,
+                               morphs.MorphXtalRDFtoPDF],
+                        qdamp=morphs.MorphResolutionDamping)
+
+
+def morph_default_config():
```
pavoljuhas: Can we rename this to `morph_config(**kw)` and process keyword arguments?
The example for `pdfmorph` could then look as follows:
```python
morph_cfg = morph_config(scale=1.01)
...
```

diffpy/pdfmorph/pdfmorph_api.py
```diff
@@ -194,21 +201,41 @@ def pdfmorph(xobj, yobj, xref, yref, rmin=None, rmax=None, rstep=None,
...
     rdat, grdat = chain.xyrefout
-    if not fig:
-        fig, ax = plt.subplots(**kwargs)
-    else:
-        ax = fig.axes[0]
-    ax.plot(rfit, grfit, label='objective')
-    ax.plot(rdat, grdat, label='reference')
+    l_list = []
+    l_list.append(ax.plot(rfit, grfit, label='objective', **kwargs))
+    l_list.append(ax.plot(rdat, grdat, label='reference', **kwargs))
```
pavoljuhas: `ax.plot` returns a _list_ of Line2D objects.  Please correct to
```
l_list = ax.plot(...) + ax.plot(...)
```

diffpy/pdfmorph/pdfmorph_api.py
```diff
@@ -0,0 +1,241 @@
...
+        Additional keyword arguements will be passed 
+        to ``ax.plot(...**kwargs)``
+
+    Returns
+    -------
+    l_list: list
+        A list of ``matplotlib.lines.Line2D`` objects representing 
+        the plotted data.
+    """
+    if not ax:
```
pavoljuhas: in this case it makes no difference, but in general it is better to use
`if ax is None` to test for missing arguments.

pavoljuhas:
> Thanks for your feedback! I've pushed changes regarding comments

Thank you @chiahaoliu, almost there.  I found a few more issues for which I posted new comments.

BTW, `git diff --check master` now reports few more errors, please correct.
I very much recommend to activate pre-commit hook to catch space issues:
```sh
cd .git/hooks
cp -i pre-commit.sample pre-commit
```

chiahaoliu:
@pavoljuhas I made more fixes regarding trailing spaces and tests. So far I am checking with `git diff --check master` and hopefully I was careful enough in squashing `pyflakes` issues.

diffpy/pdfmorph/tests/test_morph_func.py
```diff
@@ -3,6 +3,31 @@
...
+def test_morphfunc_verbose():
+    lb, ub = 1, 2
+    xref = np.arange(0.01, 5, 0.01)
+    yref = heaviside(xref, lb, ub)
+    # expand 30%
+    stretch = 0.3
+    xobj = xref.copy()
+    yobj = heaviside(xref, lb * (1 + stretch), ub * (1 + stretch))
+    cfg = morph_default_config(stretch=0.1) # off init
+    morph_rv = pdfmorph(xobj, yobj, xref, yref, verbose=True, **cfg)
```
pavoljuhas: Is there some assertion that these tests verify?

pavoljuhas:
@chiahaoliu - looks better, however travis still fails with SyntaxError due to f-strings.

diffpy/pdfmorph/tests/test_morph_func.py
chiahaoliu: It is a smoke test. I was trying to test if the verbose option inside `pdfmorph` function works. Inspired by the f-string problem, I feel it might not be a bad idea to test against py2 and older python :p

chiahaoliu:
@pavoljuhas I've fixed more f-string problems

.travis.yml
```diff
@@ -13,6 +13,8 @@ python:
 
 before_install:
   - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+  - sleep 3 # give xvfb some time to start
```
pavoljuhas: If this is for matplotlib, you can do without starting X-server:
```yaml
# use matplotlib backend that does not need DISPLAY
- export MPLBACKEND=agg
```

chiahaoliu: @pavoljuhas Thanks! It was for matplotlib and I agree suppressing x-server might be a better idea. I've included this change.

chiahaoliu:
@sbillinge @CJ-Wright maybe you could take another look before a go?

pavoljuhas:
@sbillinge - are you OK with merging this in?

diffpy/pdfmorph/pdfmorph_api.py
```diff
@@ -0,0 +1,262 @@
...
+         to ``ax.plot(...**kwargs)``
+
+    Returns
+    -------
+    l_list: list
+        A list of ``matplotlib.lines.Line2D`` objects representing
+         the plotted data.
+    """
+    if ax is None:
+        ax = plt.gca()
```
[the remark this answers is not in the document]
pavoljuhas: > Do we want this or should we build a fig, ax?

I think it is better as is because the plot can be placed to any existing axis or subplot.
The plotting function should do as little as possible to produce the plot; creating (and possibly styling) new figure is an extra task.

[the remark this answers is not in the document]
CJ-Wright: In that case we should require an axis. The getting of the current axis could cause problems with existing axes. For reference here is how this is done in bluesky code: https://github.com/bluesky/bluesky/blob/master/bluesky/callbacks/mpl_plotting.py#L92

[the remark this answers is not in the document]
pavoljuhas: > For reference here is how this is done in bluesky code ...

That opens a new figure when `ax` is not given.  I am OK with such change.

diffpy/pdfmorph/pdfmorph_api.py
```diff
@@ -0,0 +1,262 @@
...
+        output += "\n# Pearson = %f" % pcc
+        print(output)
+
+    rv_dict = dict(morph_chain=chain, morphed_config=rv_cfg,
+                   rw=rw, pcc=pcc)
+    return rv_dict
+
+
+def plot_morph(chain, ax=None, **kwargs):
+    """function to plot pdfmorph result from a morph chain
```
pavoljuhas: Let's be more direct as what this function does:
```
Plot the morphed PDF together with a reference PDF that it fits.

Open a new figure unless a specific axis is provided for plotting.
```

diffpy/pdfmorph/pdfmorph_api.py
```diff
@@ -0,0 +1,262 @@
...
+    -------
+    l_list: list
+        A list of ``matplotlib.lines.Line2D`` objects representing
+         the plotted data.
+    """
+    if ax is None:
+        ax = plt.gca()
+    rfit, grfit = chain.xyobjout
+    rdat, grdat = chain.xyrefout
+    l_list = ax.plot(rfit, grfit, label='objective', **kwargs)
```
pavoljuhas: the "objective" label is confusing - let's replace with "morph".

sbillinge:
If possible I would like to adopt language "target" and "morphed" as this
seems to be unambiguous to me.  There is a target spectrum and there is a
second spectrum that gets morphed to the target one.

I haven't looked through the code to see where, if anywhere, this is an
issue, but I think this is a good opportunity for (maybe a new issue)
cleaning this up.

S

chiahaoliu:
Yes, the language of `reference` and `objective` was inherited from previous design and I found it a bit confusing when I first started.

 Changing the naming might be a good idea. I have created another issue about this.

pavoljuhas:
@sbillinge - can we merge this?  I have some new branch that depends on fixups here.

sbillinge:
Thanks @chiahaoliu thanks @pavoljuhas

PR #20, opened 2019-06-21
pyflakes fixups

pavoljuhas:
Fix the most egregious warnings from pyflakes.
We should refactor all star imports, but that is for another PR.


pavoljuhas:
@chiahaoliu @CJ-Wright - are you OK with merging this in?

chiahaoliu:
@pavoljuhas Yes, this looks fine to me. I can do the honor too.

PR #33, opened 2019-07-15
Docs pdfmorph

RomanMatthew:
only opening this PR because I'm pretty sure I royally hiccuped on github and now have all these commits in this PR which was not supposed to be the case. @CJ-Wright and @chiahaoliu, could you guys help me out please?


RomanMatthew:
I readressed this with #34 after consulting with Tim to get rid of unnecessary commits

Issue #1, opened 2018-01-07
DOC: needs docs


CJ-Wright:
We need some docs:
- [ ] change docs over to numpy
- [x] setup sphinx
- [x] setup doctr
- [x] setup description

RomanMatthew:
I'll start working on this soon! What do you think would be the closest thing that I could model it after?

CJ-Wright:
the bluesky docs aren't bad: https://github.com/bluesky/bluesky

RomanMatthew:
@sbillinge @CJ-Wright @cfarrow @chiahaoliu @pavoljuhas 

Hello all, 

I'll use this issue to ask some content-related stuff during the doc development. Some of the blanks I'll need help filling in will be evident by a lack of detail/words, but I want to be proactive. Modeling off of the [diffpy.structure](https://diffpystructure.readthedocs.io/en/stable/) docs, my first question is whether or not you guys wanted to put anything in the "Acknowledgements" section of the front page. Let me know! I'll be rolling out more questions as development continues

sbillinge:
nice job @romanmatthew, this is very much appreciated.   We will respond as
best as we can to your questions.

PR #34, opened 2019-07-15
Added framework for online docs

RomanMatthew:
address #1 - still need some help


RomanMatthew:
@CJ-Wright I noticed that the documentation written into the code was of the same style as previous python packages we've released and that there should be some way to easily pull that into the docs, but I think I'm missing something from my conf.py file to allow that to happen. Tips?

CJ-Wright:
You might look at a conf.py here https://github.com/xpdAcq/xpdAn/blob/master/docs/conf.py

RomanMatthew:
@CJ-Wright I'm working on the conf.py file based on the one you presented and hoping it renders the text like hopefully it should. I have to run to a meeting right now but could you review my next commit to see if the `os` stuff is done right? That's what I'm most shaky about

RomanMatthew:
@CJ-Wright It seems like the problem now is the module importation but I really can't figure out what I'm doing wrong. Could you weigh in when you could? Or perhaps @pavoljuhas if you get the chance?

CJ-Wright:
Don't worry about the template stuff (line 96 down) you can remove that.

RomanMatthew:
@CJ-Wright @pavoljuhas this is, I think, my last main problem - I don't quite know how to make the system "find" the documentation written into the python files (which, to my understanding, is where the meat of the docs come from). Can you help me figure it out? 

Screen Shot 2019-07-18 at 1 58 19 PM

pavoljuhas:
> I think, my last main problem - I don't quite know how to make the system "find" the documentation written into the python files

If you use the sphinx `autodoc` extension, the package modules must be importable from the sphinx process (sphinx itself is a Python app).  This can be accomplished by adding to `sys.path` in conf.py as for example [here](https://github.com/diffpy/diffpy.structure/blob/df264a84139bc834424cc323d1bcfc1f9a4ca6cc/doc/manual/source/conf.py#L25) or another option is to run `python setup.py develop` so your source tree is visible to Python.  (assuming `which python` is the same interpreter as the one for sphinx-build)

The first approach is a bit safer, because it will find Python sources from your repo first rather than any pre-existing installation of diffpy.pdfmorph.

RomanMatthew:
@pavoljuhas @sbillinge Do you think you could give it a look now? If you can preview the docs, that is! Got the text up and working and looking mostly good, with a few more things I'd like to add once Simon and I finish hashing out the tutorial!

sbillinge:
@RomanMatthew pls can you send me instructions how to build the docs?

I cloned your branch and I am running ``make html`` in the directory with the Makefile in it but I am getting a moduleNotFound error:  ``ModuleNotFoundError: No module named 'sphinx_rtd_theme'``

What commands are you using to build the docs?

doc/source/index.rst
```diff
@@ -0,0 +1,70 @@
...
+Finally, we note that though ``PDFmorph`` should work on other spectra 
+that are not PDFs, it has not been extensively tested beyond the PDF.
+
+
+Authors
+-------
+
+``PDFmorph`` is developed by members of the Billinge Group at 
+Columbia University and Brookhaven National Laboratory including 
+Pavol Juhás, Christopher L. Farrow, Christopher Wright, Chia-Hao
```
sbillinge: Chris Farrow first, please fix parens in Tim's name

doc/source/conf.py
```diff
@@ -0,0 +1,95 @@
...
+#
+import sys
+import os
+sys.path.insert(0, os.path.abspath('../..'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'PDFmorph'
+copyright = '2019, Chris Farrow, Pavol Juhás, Chia-Hao (Timothy) Liu, S. Matthew Roman, Simon J.L. Billinge'
```
sbillinge: copyright is Columbia.  Please get it from the license file.

doc/source/index.rst
CJ-Wright:
```suggestion
Pavol Juhás, Christopher L. Farrow, Christopher J. Wright, Chia-Hao 
```

Issue #27, opened 2019-07-09
Package PDFmorph tutorial files


RomanMatthew:
Make HTML and PDF manual modeled after [PDFgui HTML manual](https://www.diffpy.org/doc/pdfgui/index.html) and [PDFgui PDF manual](https://www.diffpy.org/doc/pdfgui/pdfgui.pdf) with sources [here](https://github.com/diffpy/diffpy.pdfgui/tree/master/doc/manual)

PR #36, opened 2019-07-16
Added tutorial to homepage of git repo

RomanMatthew:
contribute to #27


RomanMatthew:
@sbillinge @chiahaoliu Can you let me know if this justifies closing #27 ?

sbillinge:
I am not sure I understand your question.  I think that this PR closes #27 when it is merged.

Is it ready for review?  Who have you tested the tutorial on?

TUTORIAL.rst
```diff
@@ -0,0 +1,59 @@
...
+
+	pdfmorph --help
+
+a list of additional modifications that you can run in your PDFmorph
+program is generated. 
+
+You can add any of these "options" by placing them before your two 
+filenames. ::
+
+	pdfmorph [options] <testfile 1> <testfile 2>
```
sbillinge: honestly, I think I prefer introducing these things in a step-by-step way.  You can use rhetorical tricks like

type ``pdfmorph --scale=0.5 ni.gr``

It is also a good idea to have them do some things wrong, so if tehy try and morph a higher temperature dataset onto a lower one, what happens?

RomanMatthew: Ok, I'll try to make it more step-by-step next time around (currently trying to figure out the style sheet). So far, I've been using PDF data provided in the repo (tests/testdata) so I don't really know which are higher and lower temps at this point. Could you send me some data that's more clearly marked for temp so I can see for myself too?

TUTORIAL.rst
```diff
@@ -0,0 +1,59 @@
...
+before using this application.
+
+After all of the necessary software is installed, activate your PDFmorph-
+equipped conda environment and change into a directory containing at
+least two PDF files. If you need some test files to help you get started,
+check `here<https://github.com/diffpy/diffpy.pdfmorph/tree/master/tests/testdata>`_.
+
+Once you have two PDF files, run the command ::
+
+	pdfmorph <testfile 1> <testfile 2>
```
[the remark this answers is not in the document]
RomanMatthew: Should I encourage users to download the example files on the github, or provide different example files?

TUTORIAL.rst
```diff
@@ -0,0 +1,59 @@
...
+You can add any of these "options" by placing them before your two 
+filenames. ::
+
+	pdfmorph [options] <testfile 1> <testfile 2>
+
+For example, running the command ::
+
+	pdfmorph --scale=0.5 --smear=0.5 <testfile 1> <testfile 2>
+
+will scale testfile 1 by a factor of 0.5 and will smear its peaks with a
```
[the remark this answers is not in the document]
RomanMatthew: To make sure I understand it correctly:

When the program is fitting the "smear" parameter, it's broadening the morphed PDF's peaks to minimize the difference between the target and itself. When the program is smearing with a fixed  "smear" parameter, it's broadening the morphed PDF's by some arbitrary factor. Is that right? And same with scale and the other options as well, right?

sbillinge:
That's correct Matt, though it is not an "arbitrary" number it is a
user-supplied number.

RomanMatthew:
@sbillinge Please give it a look when you can!

RomanMatthew:
Once this is ready I should also be able to add it to the doc and publish those on rtd!

TUTORIAL.rst
```diff
@@ -0,0 +1,112 @@
...
+
+Before you've started this tutorial, please ensure that you've installed
+all necessary software and dependencies.
+
+PDFmorph Example
+----------------
+
+	1. Open your Terminal or Command Prompt.
+
+	2. It it's not active already, activate your PDFmorph-equipped
```
sbillinge: If

TUTORIAL.rst
```diff
@@ -0,0 +1,112 @@
...
+----------------
+
+	1. Open your Terminal or Command Prompt.
+
+	2. It it's not active already, activate your PDFmorph-equipped 
+	   conda environment by typing in ::
+	
+		source activate <pdfmorph_env>
+
+	   on Linux or ``activate <pdfmorph_env>`` on Windows.
```
sbillinge: "where ``<pdfmorph_env>`` is the name of your environment."

btw, you just need this once and people will get it I think....

TUTORIAL.rst
```diff
@@ -0,0 +1,112 @@
...
+		* If you need to list your available conda environments,
+		  run the command ``conda info --envs`` or 
+		  ``conda env list``
+
+	3. Change directories with the ``cd`` command to a directory
+	   containing PDF files. For instance, I will use the file
+	   "ni_qmax25.cgr", found `here<https://github.com/diffpy/diffpy.pdfmorph/tree/master/tests/testdata>`_,
+	   so I will run ``cd ~/diffpy.pdfmorph/diffpy/pdfmorph/tests/testdata``
+
+	4. Using this file for an example (you can download it from the
```
sbillinge: Rather than write it like this I would say before.

"In this tutorial we will use test data suppled as an example in the PDFmorph package

1. download the data by typing ....
1. move it to a directory where you want to work, e.g.,.....
1. cd to that directory....

and so on, so don't make it general, make it specific.

Then take a moment to describe what is in that folder.   .cgr files are calculated PDFs from PDFgui, .gr files are measured PDFs, then some words about what was measured, where it was measured, under what conditions and what temperatures etc.  Also acknowledge the experimenters if there is no paper reference.

TUTORIAL.rst
```diff
@@ -0,0 +1,112 @@
...
+	   on Linux or ``activate <pdfmorph_env>`` on Windows.
+
+		* If you need to list your available conda environments,
+		  run the command ``conda info --envs`` or 
+		  ``conda env list``
+
+	3. Change directories with the ``cd`` command to a directory
+	   containing PDF files. For instance, I will use the file
+	   "ni_qmax25.cgr", found `here<https://github.com/diffpy/diffpy.pdfmorph/tree/master/tests/testdata>`_,
+	   so I will run ``cd ~/diffpy.pdfmorph/diffpy/pdfmorph/tests/testdata``
```
sbillinge: I will type ``cd ...``

TUTORIAL.rst
```diff
@@ -0,0 +1,112 @@
...
+
```
sbillinge: The first thing to have them do is to type ``PDFmorph -h`` so they know how to access the help.

Also, at this point, encourage them to joing diffpy-users and let them know they can post questions about how to use the software (also bug reports) there.

TUTORIAL.rst
```diff
@@ -0,0 +1,112 @@
...
+		.. figure:: doc/images/morph_ex1.png
+			:align: center
+
+	   A few things to note:
+
+		* The green line (the difference curve) is flat because
+		  in this example, we used the same PDF twice (so, there
+		  is no difference)
+
+		* The red curve is the taget PDF, and the blue circles
```
sbillinge: We usually use red line for model and blue dots for data (and green line offset for difference) in our normal PDF papers where we fit models to data.  In the PDFmorph context they are usually both data, so we should think about what makjes the most sense, but if we stick with red line and blue circles, the target should be the blue and the morph should be the red.

For the figure, ticks inside the box please.  Also, no legend, though reproducing the Rw in the panel I think is a would be a nice touch.

TUTORIAL.rst
```diff
@@ -0,0 +1,112 @@
...
+		  represent the morphed PDF; this will become clearer 
+		  with the next example.
+
+	5. Now, using the files "ni_qmax25_e17.5_p5.0.cgr" as our target
```
sbillinge: change all the filenames to be more meaningful to a novice user

TUTORIAL.rst
```diff
@@ -0,0 +1,112 @@
...
+	   and "ni_qmax25_psize30.cgr" as our morphed PDF (available 
+	   from the same source as above), run the following to see the 
+	   program's "morphing" capabilities ::
```
sbillinge: it seems like for them to see the morphing capabilities they should see what they look like without being morphed, and compare it to what they look like after morphing?

TUTORIAL.rst
sbillinge: I would start with a few words about what PDFmorph is doing (more or less cut and paste from the README), then type, "in general you run PDFmorph with the following command ``pdfmorph <target PDF> <morph PDF>`` "(check I have it the right way around)  " what happened here is we made the same PDF file to be target and morph ..."  tell the story.

TUTORIAL.rst
```diff
@@ -0,0 +1,112 @@
...
+
+		pdfmorph ni_qmax25_e17.5_p5.0.cgr ni_qmax25_psize30.cgr
+
+	   This should produce an image like this ::
```
sbillinge: This seems to be a very bad example because the (default) morphing was not successful.  These data may have been in there to illustrate a particle-size effect morph

TUTORIAL.rst
```diff
@@ -0,0 +1,112 @@
...
+
+	   This should produce an image like this :: 
+
+		.. figure:: doc/images/morph_ex2.png
+			:align: center
+
+	   Note how the difference curve is now showing the difference
+	   between the two PDFs, after the program has "morphed" the 
+	   second PDF to the target PDF.
+
```
sbillinge: need more different examples.  What happens when you turn off a morphing action, then introduce them one at a time? what happens when you fix morphing parameters to user-defined values.  You probably need to work with someone who knows PDF to get the story right because there are lots of "what if you have data where one has a smaller particle size?" kind of scenarios that you are testing for.

TUTORIAL.rst
```diff
@@ -0,0 +1,112 @@
...
+inputs these values, it overrides the "choice" that the program would 
+make.
+
+Using the list provided by running ``pdfmorph --help``, you should now 
+be able to "morph" your PDFs in a number of highly customizable ways.
+
+Bug Reports
+-----------
+
+Please enjoy using our software! If you come accross any bugs in the
```
sbillinge: I think it is maybe better to have them report them to diffpy-users Google group.

TUTORIAL.rst
RomanMatthew: Can you tell me if this same issue happens for you? When I branched out form origin/master, I didn't see any available "tests" folder in `diffpy.pdfmorph/diffpy/pdfmorph`. That's the only reason I was hesitant of including this.

RomanMatthew: Maybe it would be better to include some different, more clearly documented pdf data in a different folder called "tutorialdata" or something? I don't have any PDF data on hand and I'm not sure that it would be okay to use some off the internet, and I can't discern the information you're describing in the files we have in the repo already. Would it be possible for someone to merge a tutorialdata dir into the homepage of the repo with some .gr files and let me know the info so I can include that? Also, is there some better alternative to having them download the specific file without needing to clone the whole repository? Sorry for all the questions; I want to come up with a good solution!

RomanMatthew:
I think a lot of the issues you brought up are really valid and could be addressed effectively if I had some palatable PDF data to work with. If I could get some as I described in my comment, I could address these ASAP!

sbillinge:
You will have to work with one of the guys to get some representative data,
and to come up with a good story   Soham is good at this stuff, but won't
be in next week so maybe just ask around and see who you get a response
from.  It might be good to know what representative datasets you need.
Maybe reach out to Rob Koch at BNL?

To illustrate things properly you need something that goes through a phase
transition and you want 4 temperature points, 2 above and 2 below.  You can
show that you can successfully morph the 2 above the transition onto each
other, and also the two below the transition, but not any that go across
the transition.

If we want to illustrate the particle form factor morph you will need two
more datasets that are non-nanoparticle and nanoparticle of the same
material.

How do we get the data to the people?  One way is of course they are
somewhere on the web and they can download them.  I am reluctant to send
them to GitHub to get them, though we could (for sure we should store them
at GitHub even if we don't send people there, for archival reasons) because
this will be a bit confusing and unfriendly an interface for non programmer
people.  Is it possible to upload files to readthedocs or linked to
github.io or wherever the docs are?  That could be an option.

Another thought is if they can be included in the package that comes
through Conda, then they will be installed in the Anaconda tree.  It is
also ugly to send them because it will be a really long and user-specific
file path, not easy to find.

Another possibility to look into is that they are installed with the
software, and we include functionality such that if you type ``PDFmorph
--install_tutorial`` a function will run that will go and get teh files
from the Anaconda tree and copy them to the current working directory.

There may be another way that CJ knows about.

S

RomanMatthew:
hey @sbillinge just wanted to push a quick reminder to review this when you get the chance!

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+		  ``conda env list``
+
+        * Run the ``pdfmorph --help`` command and read over the
+          info on that page for a brief overview of some of what we will
+          explore in this tutorial.
+
+	3. Using the ``mkdir`` command, create a directory where you'll
+        store the tutorial PDF files and use the ``cd`` command to change
+        into that directory. You can download the tutorial files
+        :download:`here <tests/testdata/tutorialData.zip`. Then, ``cd``
```
sbillinge: will this work?  It is not a URL but a relative path to some local file.  It assumes that the program is being run in a particular directory which is unlikely to be the case.

Presumably the best is if this link points to a URL?

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+        into the ``tutorialData`` directory.
+
+        * The files in this dataset were collected by Soham Banerjee
```
sbillinge: Soham is at Columbia.  The data were collected at BNL, but put the whole attribution including the beamline

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+          at Brookhaven National Laboratory in Upton, New York.
+
+        * The files are PDF data collected on Iridium Telluride with
```
sbillinge: are we going to use all 44 files in the tutorial?  It seems like we only need a handful of files for this tutorial

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+          20% Rhodium Doping (IrRhTe2) with the first file (01) collected
+          at 10K and the last (44) at 300K. The samples increase in
+          temperature as their numbers increase. The "C" in their names
+          indicates that they have undergone cooling.
```
sbillinge: This requires more explanation.  What does "undergone cooling" mean.

Has this been reviewed by @sbanerjee.  This should happen before me.....

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+
+        * Note that these files have the ``.gr`` extension, which
+          indicates that they are measured PDFs. The ``.cgr`` file
+          extension indicates that a file is a calculated PDF, sich as
```
sbillinge: typo.  Please try and avoid sending things for review with typos.

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+          those generated by the `PDFgui <https://www.diffpy.org/products/pdfgui.html>`_
+          program.
+
+    4. First, we will run the PDFmorph application without any morphing
```
sbillinge: This would be better if it was more of a story.  You can make the story up.  Soham can help.

"Let's assume you are a student and have measured these samples as a function of temperature and you want to know if they have gone through a phase transition and changed structure.  A way to do this is to compare the PDFs  measured at two different temperatures.  To do this we would like to plot them on top of each other with a difference curve plotted offset below.  PDFmorph can do that.  type
```
pdfmorph --no-morph filename_100K.gr filename_150K.gr
```
"

Something like that.

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+       and only using one PDF. Type the following command into your
+       Command Line ::
+
+            pdfmorph darkSub_rh20_C_01.gr darkSub_rh20_C_01.gr
```
sbillinge: change the file names to something more meaningful.

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+
+        This should produce two PDF curves which are congruent, resulting
+        in a flat green line underneath them.
+
+    5. Now, we will see PDFmorph run with two different PDFs and no
```
sbillinge: This should be first.  It is just plotting them which is the first thing that people would do (see the story above).

The workflow is you plot them, you see differences, you ask, are the differences due to a structure-change or just boring thermal expansion?, you morph. (a) you get a flat line after morphing (themal expansion) (b) you don't (phase transition).  So this needs to come through in the story.

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+       morphing. Type the following command into your command line ::
+
+            pdfmorph darkSub_rh20_C_01.gr darkSub_rh20_C_44.gr
+
+       Without morphing, the difference Rw = 0.407. This indicates that
+       the two PDFs vary drastically.
+
+            * While running the ``pdfmorph`` command, it is important
```
sbillinge: please double-check this is right.

The best way to get this across is probably to reiterate here that the command is 
`` pdfmorph <morphed PDF file> <target PDF file> ``

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+              to remember that the first PDF file argument you provide
+              (in this case, ``darkSub_rh20_C_01.gr``) is the PDF which
+              will get morphed, while the second PDF file argument you
+              provide (here, ``darkSub_rh20_C_44.gr``) is the PDF which
+              acts as the model and does not get morphed. Hereinafter,
+              we will refer to the first PDF argument as the "objective"
```
sbillinge: ugh, I am so over this terminology....we are using ``target`` and ``morphed``

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+              and the second as the "reference", as the PDFmorph display
+              does.
+
+    6. Now, we will start the morphing process, which requires us to
+       provide initial guesses for our scaling factor, Gaussian smear,
+       and stretch, separately. We will start with the scaling factor.
+       Begin by typing the command ::
+
+            pdfmorph --scale=2 -a darkSub_rh20_C_01.gr darkSub_rh20_C_44.gr
```
sbillinge: the ``-a`` switch applies the transformation but doesn't refine it so the language of "initial guess" is not right here

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+
+            pdfmorph --scale=0.9 -a darkSub_rh20_C_01.gr darkSub_rh20_C_44.gr
+
+       The difference Rw is now 0.351, lower than our unmorphed
+       example's value. To see PDFmorph optimize the scale factor,
+       simply drop ``-a`` from the command and type ::
+
+            pdfmorph --scale=0.9 darkSub_rh20_C_01.gr darkSub_rh20_C_44.gr
+
+       PDFmorph, given a reasonable initial guess, will use find the
```
sbillinge: use

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+       optimal value for each morphing feature. Here, we see that
+       PDFmorph displays ``scale = 0.799025`` in the command prompt,
```
sbillinge: I am pretty sure that PDFmorph will successfully converge with the default initial guess, or even with the bad initial guess of 2 (not sure about that....did you test it?) so this may not be a great example of introducing the concept of the initial guess.  The default use of the program should be that you don't give it initial guesses unless it doesn't converge.

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+              does.
+
+    6. Now, we will start the morphing process, which requires us to
+       provide initial guesses for our scaling factor, Gaussian smear,
+       and stretch, separately. We will start with the scaling factor.
+       Begin by typing the command ::
+
+            pdfmorph --scale=2 -a darkSub_rh20_C_01.gr darkSub_rh20_C_44.gr
+
+       Now, the difference Rw = 1.457, a significant increase from our
```
sbillinge: the tutorial will be better if you insert figures showing the datasets plotted on top of each other (i.e., the output from PDFmorph given the inputs you gave.  This will be sanity check for the user that they are gietting what they should be getting

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+
+            pdfmorph --scale=0.8 --smear=0.5 -a darkSub_rh20_C_01.gr darkSub_rh20_C_44.gr
+
+       And viewing the results. We've tailored our scale factor to be
+       close to the value given by PDFmorph, but see that the difference
+       Rw has increased substantially due to our smear value. One
+       approach, as described above, is to remove the ``-a`` from the
+       above command and run it again.
+
+            * Note: The warnings that the Terminal/Command Prompt
```
sbillinge: in which case should be removed from the code before releasing.....thanks for catching that.

TUTORIAL.rst
```diff
@@ -0,0 +1,187 @@
...
+              displays are largely numerical in nature and do not
+              indicate a physically irrelevant guess. These are somewhat
+              superficial and in most cases can be ignored.
+
+       We see that this has had hardly any effect on our PDF. To see
```
sbillinge: we can probably find a more interesting story for introducing this the rmin rmax values

PR #43, opened 2019-07-18
Fix setup metadata

pavoljuhas:
## TODO

- [x] update `maintainer`
- [ ] fix date value in version.py


pavoljuhas:
I think we should update the maintainer unless @cfarrow would like to hold on to this.
@sbillinge - any suggestions?

sbillinge:
we can put me maybe.  I can always delegate it.

pavoljuhas:
@CJ-Wright - can we get rever to update the `__version__` and `__date__` values in
[version.py](https://github.com/pavoljuhas/diffpy.pdfmorph/blob/5e2ca6e3dbeff3fd09c9781e4cb172aedff866f8/diffpy/pdfmorph/version.py#L22-L26)?

Can we get the `__git_commit__` value to version.py with rever?

Can it create gpg-signed tag in the release action?

CJ-Wright:
Most likely yes to all three. You can write your own activities which perform those operations rather than using the stock ones if you want. See the custom activities for Google cloud things here https://github.com/regro/conda-metachannel/blob/master/rever.xsh

CJ-Wright:
I'm sure some of these would be welcome additions to the stock rever activities.

PR #50, opened 2022-11-16
WIP changed the labels to be the input file names and switched the colors…

### LINKED ISSUE #16: Replace style code with stylesheet

### LINKED ISSUE #46: Give file names instead of "reference" and "objective" in the plots

chloeann95:
… so that the 1st pdf would be a blue line and 2nd red circles to correspond with temperature. 
closes #46 
closes #16


diffpy/pdfmorph/pdfmorphapp.py
```diff
@@ -254,7 +254,8 @@ def main():
 
     if opts.plot:
         pairlist = [chain.xyobjout, chain.xyrefout]
-        labels = ["objective", "reference"]
+        labels = [pargs[0], pargs[1]]
+        print(pargs[0], pargs[1])
```
sbillinge, line 258: please remove.  I guess this was a debugging line

sbillinge (commented):
Thanks Chloe, this is really great!

Please see my inline comment.  Our workflow is that you always answer inline comments.  Simply put "done" if you do it, or whatever comment you have. But if you don't respond to the comments I don't know if you saw them or what you thought, even though you do because it is in your head....if that makes sense.

We will need a "news" item for the rever release. We do one of these for each branch and this gets made into the changelog so make it descriptive and public facing.  The workflow is to copy TEMPLATE.rst to <branch_name>.rst (in the same directory as TEMPLATE) and then fill in what has changed or added.  This would be a change I think.

Also, we were going to add the billinge-group style sheet so that needs to be imported and added to the requirements as a dependency.  I am not sure if code needs to be updated to make it work, but maybe the blue or red need to be removed or changed.

Before I merge, could you share screen shots of what the plots look like and we can do all the tweaks to make it as pretty as possible.

diffpy/pdfmorph/pdfmorphapp.py
chloeann95: Which template are you referring to here? I have the template files from the paper but is there something else? Also, where is the style sheet located? Is it the same as the group meeting one? Also, yes, that was a debugging line will remove it before doing the next PR.

sbillinge: This is standard rever release workflow.  Please refer to the regro/rever documentation and/or talk to songsheng.  You can see examples in  the regolith repo and there may be something in the wiki on this, I am not sure.  I presume you looked there already.

I don't see a comment on the style-sheet.  Can you figure that one out?

chloeann95: Alright will do. For the style-sheet I just wanted to check where it was located at, I can figure out the rest myself. Thanks

sbillinge: sound good. Please check the wiki and get back if you can't find info there on the style-sheet.

sbillinge (commented):
please see inline, and comments in slack about importing the stylesheet from conda.

PR #54, opened 2023-02-04
added bg-style-sheet to replace existing plotting

### LINKED ISSUE #16: Replace style code with stylesheet

chloeann95:
bg-style-example

Here's an example of what the stylesheet plotting looks like. I left everything that was there before commented out and can trial and error what we do or don't want.

Closes #16

### DIFF
--- a/diffpy/pdfmorph/pdfmorphapp.py
+++ b/diffpy/pdfmorph/pdfmorphapp.py
@@ -26,6 +26,8 @@ import diffpy.pdfmorph.pdfplot as pdfplot
 import diffpy.pdfmorph.morphs as morphs
 import diffpy.pdfmorph.refine as refine
 
+from bg_mpl_stylesheet.bg_mpl_stylesheet import bg_mpl_style
+import matplotlib as plt
 
 def createOptionParser():
 
@@ -114,6 +116,7 @@ radius RADIUS and polar radius PRADIUS. If only PRADIUS is specified, instead ap
     return parser
 
 def main():
+    plt.style.use(bg_mpl_style)
     parser = createOptionParser()
     (opts, pargs) = parser.parse_args()
 
--- a/diffpy/pdfmorph/pdfmorphapp.py
+++ b/diffpy/pdfmorph/pdfmorphapp.py
@@ -26,8 +26,6 @@ import diffpy.pdfmorph.pdfplot as pdfplot
 import diffpy.pdfmorph.morphs as morphs
 import diffpy.pdfmorph.refine as refine
 
-from bg_mpl_stylesheet.bg_mpl_stylesheet import bg_mpl_style
-import matplotlib as plt
 
 def createOptionParser():
 
@@ -116,7 +114,6 @@ radius RADIUS and polar radius PRADIUS. If only PRADIUS is specified, instead ap
     return parser
 
 def main():
-    plt.style.use(bg_mpl_style)
     parser = createOptionParser()
     (opts, pargs) = parser.parse_args()
 
--- a/diffpy/pdfmorph/pdfplot.py
+++ b/diffpy/pdfmorph/pdfplot.py
@@ -53,8 +53,8 @@ def plotPDFs(pairlist, labels=None, offset ='auto', rmin = None, rmax = None):
         plt.plot(r, gr + idx * offset, label = labels(fit))
     plt.xlim(rmin, rmax)
 
-    #if gap == 0:
-    #    plt.legend(loc = 0)
+    if gap == 0:
+        plt.legend(loc = 0)
 
     plt.legend()
     plt.xlabel(r"$r (\mathrm{\AA})$")
@@ -133,9 +133,6 @@ def comparePDFs(pairlist, labels=None, rmin = None, rmax = None, show = True,
     ymax = max(diff[sel])
     offset = -1.1*(ymax - ymin)
 
-    # Set up the plot
-    #_configure()
-
     # Scale the x-limit based on the r-extent of the signal. This gives a nice
     # density of PDF peaks.
     rlim = rvmax - rvmin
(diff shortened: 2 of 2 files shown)


diffpy/pdfmorph/pdfmorphapp.py
```diff
@@ -114,6 +116,7 @@ def createOptionParser():
     return parser
 
 def main():
+    plt.style.use(bg_mpl_style)
```
sbillinge, line 119: Is this needed here, or only where it is used in the plotting module?  Delete here if it is not used (and delete the import)

diffpy/pdfmorph/pdfplot.py
```diff
@@ -217,16 +222,17 @@ def truncatePDFs(r, gr, rmin = None, rmax = None):
...
-    plt.rc("axes", linewidth = 2, labelsize = 30)
-    plt.rc("xtick", labelsize = 25)
-    plt.rc("xtick.major", size = 7)
-    plt.rc("xtick.minor", size = 3)
-    plt.rc("ytick", labelsize = 25)
-    plt.rc("ytick.major", size = 7)
-    plt.rc("ytick.minor", size = 3)
-    plt.rc("legend", fontsize = 18)
-    plt.rc("lines", markeredgewidth = 2) # thicker axes and symbols
+    plt.style.use(bg_mpl_style)
```
sbillinge, line 225: remove?

diffpy/pdfmorph/pdfplot.py
```diff
@@ -133,32 +138,32 @@ def comparePDFs(pairlist, labels=None, rmin = None, rmax = None, show = True,
 
     # Scale the x-limit based on the r-extent of the signal. This gives a nice
     # density of PDF peaks.
-    rlim = rvmax - rvmin
-    scale = rlim / 25.0
+    #rlim = rvmax - rvmin
```
sbillinge, line 141: are you sure you need to remove these?  I am not 100% sure but don't they scale the axes to the range of the data, something like that?  Just check.

Go ahead and delete all the commented lines when you are sure they are not needed.  Do this on a single commit (i.e. there are no other changes in that commit) which would help us to get them back if we needed them again for some reason.

diffpy/pdfmorph/pdfplot.py
```diff
@@ -133,32 +138,32 @@ def comparePDFs(pairlist, labels=None, rmin = None, rmax = None, show = True,
...
-    plt.plot(rfit, grfit, label = labelfit, linestyle = 'solid', linewidth =
-            2, color = 'red')
-    plt.plot(rdat, offset*numpy.ones_like(diff), linestyle = '--', linewidth
-            = 1, color = 'black', dashes = (15, 15), aa = False)
+    plt.plot(rdat, grdat, label = labeldata) #, marker = 'o', markerfacecolor
+    #        = 'white', markeredgecolor = 'blue', markersize = 7,
+    #        markeredgewidth = 0.75)
+    plt.plot(rfit, grfit, label = labelfit) #, linestyle = 'solid', linewidth =
+    #        2, color = 'red')
+    plt.plot(rdat, offset*numpy.ones_like(diff)) #, linestyle = '--', linewidth
```
sbillinge, line 162: we usually like the offset diff curve to be green.  This can be accomplished easily here if we change the order of plotting the numpy.ones and the diff.   

even better, override the color just for the horizontal line to be black.  I am not sure if this will break how the color iterator works, but it should be ok if this is the last plot command.  Just try it.

diffpy/pdfmorph/pdfplot.py
```diff
@@ -217,16 +222,17 @@ def truncatePDFs(r, gr, rmin = None, rmax = None):
...
-    plt.rc("xtick", labelsize = 25)
-    plt.rc("xtick.major", size = 7)
-    plt.rc("xtick.minor", size = 3)
-    plt.rc("ytick", labelsize = 25)
-    plt.rc("ytick.major", size = 7)
-    plt.rc("ytick.minor", size = 3)
-    plt.rc("legend", fontsize = 18)
-    plt.rc("lines", markeredgewidth = 2) # thicker axes and symbols
+    plt.style.use(bg_mpl_style)
+    plt.rc("font") #, size = 40)
```
sbillinge, line 226: I think you can delete all these as they should be being handled in the style-sheet.  If there are any things we want to override (such as the legend fontsize) we can leave that in, but keep it to a minimum.  In fact, as you ahve done, switch them all off (but do it by deleting them) then only add them back in as needed on esthetic grounds.

diffpy/pdfmorph/pdfplot.py
```diff
@@ -133,32 +138,32 @@ def comparePDFs(pairlist, labels=None, rmin = None, rmax = None, show = True,
...
-    rlim = rvmax - rvmin
-    scale = rlim / 25.0
+    #rlim = rvmax - rvmin
+    #scale = rlim / 25.0
     # Set a reasonable minimum of .8 and maximum of 1
-    scale = min(1, max(scale, 0.8))
-    figsize = [13.5, 4.5]
-    figsize[0] *=  scale
-    fig = plt.figure(1, figsize = figsize)
+    #scale = min(1, max(scale, 0.8))
```
sbillinge, line 144: do you still not need

diffpy/pdfmorph/pdfplot.py
```diff
@@ -81,6 +84,8 @@ def comparePDFs(pairlist, labels=None, rmin = None, rmax = None, show = True,
     line.  The difference curve will be in green and offset for clarity.
     
     """
+    plt.style.use(bg_mpl_style)
```
sbillinge, line 87: I don't think this is needed here.   When you set the plot-style it is set globally.  Remove and see if it still works as expected.

sbillinge (commented):
good start.  Please see my comments .  Push changes to the same PR and we can continue the conversation.

sbillinge:
@chloeann95 another quick  suggestion is to make you commits a bit more granular as you fix things.  What you did here is great, np, but in general, as you edit, we may find it easier later to unwind mistakes if the commits are smaller.  For example, to try and remove the extra global style declerations, I would try removing them at different places and make sure the figure still builds the same, clean everything up, and then put this on a single commit, then move on to the next thing, and so on.  You don't have to push each commit.  You can work on multiple commits and then push when you are ready.  

Also, I would say it is a priority to set up CI.  We use GitHub actions at this point.   we can discuss at our meeting (also maybe set it up)

diffpy/pdfmorph/pdfmorphapp.py
chloeann95: You're right, I saw some of the config settings and thought it might be used but didn't look deeper because the plot was working after changing the other file. Removed.

diffpy/pdfmorph/pdfplot.py
chloeann95: Removed, thought it might be needed for each of the functions because I didn't write it the same way the wiki says to set it as the default. Tested and it works fine.

diffpy/pdfmorph/pdfplot.py
chloeann95: removed, legend was handled separately already and I reintroduced it, it was commented out previously

diffpy/pdfmorph/pdfplot.py
```diff
@@ -128,9 +133,6 @@ def comparePDFs(pairlist, labels=None, rmin = None, rmax = None, show = True,
     ymax = max(diff[sel])
     offset = -1.1*(ymax - ymin)
 
-    # Set up the plot
-    _configure()
```
sbillinge, line 132: if this function is not being used any more, we should delete it (the function, not the call to it). Again, do it on a separate commit so we can revert if necessary.

sbillinge (commented):
lookin' good.  Please can you see my one comment, and also post a new version of what the plot looks like?  We can then give it one last look and decide if we want to make any changes.

PR #67, opened 2023-03-13
Changed 'reference' and 'objective' to the file names and switched ordering to match input order

### LINKED ISSUE #46: Give file names instead of "reference" and "objective" in the plots

chloeann95:
closes #46 

issue46_before
issue46_after

### DIFF
--- a/diffpy/pdfmorph/morphs/morph.py
+++ b/diffpy/pdfmorph/morphs/morph.py
@@ -47,10 +47,12 @@ class Morph(object):
     Instance attributes:
 
     config      -- dictionary that contains all configuration variables
+    "Objective" is the 1st input file, displayed in red
     xobjin      -- last objective input x data
     yobjin      -- last objective input y data
     xobjout     -- last objective result x data
     yobjout     -- last objective result y data
+    "Reference: is the 2nd input file, displayed in blue
     xrefin      -- last reference input x data
     yrefin      -- last reference input y data
     xrefout     -- last reference result x data
@@ -120,8 +122,8 @@ class Morph(object):
     def morph(self, xobj, yobj, xref, yref):
         '''Morph arrays objective or reference.
 
-        xobj, yobj  --  Objective arrays.
-        xref, yref  --  Reference arrays.
+        xobj, yobj  --  Objective arrays (input 1).
+        xref, yref  --  Reference arrays (input 2).
 
         Identity operation.  This method should be overloaded in a derived
         class.
--- a/diffpy/pdfmorph/pdfmorphapp.py
+++ b/diffpy/pdfmorph/pdfmorphapp.py
@@ -325,8 +325,8 @@ def main():
         outfile.close()
 
     if opts.plot:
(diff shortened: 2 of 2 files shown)


diffpy/pdfmorph/morphs/morph.py
```diff
@@ -47,10 +47,12 @@ class Morph(object):
     Instance attributes:
 
     config      -- dictionary that contains all configuration variables
+    "Objective" is the 1st input file, displayed in red
     xobjin      -- last objective input x data
     yobjin      -- last objective input y data
     xobjout     -- last objective result x data
     yobjout     -- last objective result y data
+    "Reference: is the 2nd input file, displayed in blue
```
sbillinge, line 55: I think we wanted to move away from the "reference" and "objective" names as this is confusing.  Which is the target curve and which is the morphed one?   Maybe use "target" and "morphed" which seems to be much less ambiguous.

To fix this properly we also want to refactor the variable names, so `xobjin` would go to `xtargetin` and `xrefin` would go to ` xmorphedin`

diffpy/pdfmorph/morphs/morph.py
```diff
@@ -120,8 +122,8 @@ def __init__(self, config=None):
     def morph(self, xobj, yobj, xref, yref):
         '''Morph arrays objective or reference.
 
-        xobj, yobj  --  Objective arrays.
-        xref, yref  --  Reference arrays.
+        xobj, yobj  --  Objective arrays (input 1).
```
sbillinge, line 125: keep this reference to which order the arrays are added, but again change the name

sbillinge (commented):
let's see if we can use this opportunity to increase test coverage.  I think it is not so worthwhile to test matplotlib plot lines, but some of things can maybe be tested.

sbillinge (commented):
Please write tests where needed and I can merge this.  You can see all the lines not covered in the codecov report on GH.  Here are my suggestions:

- [ x] no tests needed for `plot` lines
- [ ] tests needed for `createOptionParser()` function.  We want to test the info is correctly loaded and the edge cases, so no need to run anything from the command line (we assum optparse will work) so it is more testing that the args are correclty built (e.g., see https://stackoverflow.com/questions/18160078/how-do-you-write-tests-for-the-argparse-portion-of-a-python-module)
- [ ] tests for `f` in `morphshape.py`
- [ ] if possible, test for `add_pearson` option in `pdfmorph_api`
- [ ] test for correct insertion of `bg_mpl_style` in `pdfplot.py`.  Don't test everything, because if we change a style value in the style sheet later, we want the change to propagate through presumably, so just a quick test that the right style sheet was installed if you can figure it out.  maybe there is an atribute in the stylesheet with its name or something.
- [ ] probably don't need tests for the plotting functions in `pdfplot.py`, comparePDFs and `plotPDFs` though we could remove the codecov errors by having them return something and testing that it is correctly returned.
- [ ] tests in `tools.py`
- [ ] tests in `version.py`

PR #68, opened 2023-03-13
Added input morph parameters to the output

### LINKED ISSUE #44: PDFmorph outputs the values of the parameters that it used
I may just not know how to use the software, but is there no way to output the values it used for the morphing?  Add values to the output.

chloeann95:
closes #44 

Here's before
Screen Shot 2023-03-13 at 4 50 54 AM

and after
Screen Shot 2023-03-13 at 4 50 29 AM

and here is the new save file that it creates for the morphed PDF in case we should avoid altering the file format I can revert the additional parameters in the save file.
Screen Shot 2023-03-13 at 4 53 56 AM

### DIFF
--- a/diffpy/pdfmorph/pdfmorphapp.py
+++ b/diffpy/pdfmorph/pdfmorphapp.py
@@ -199,6 +199,9 @@ def main():
     xref, yref = getPDFFromFile(pargs[1])
 
     # Get configuration values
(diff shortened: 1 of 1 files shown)


chloeann95:
just noticed I need to include #'s in front of the new text outputs

PR #69, opened 2023-03-13
Renamed top level tests folders to tutorials

### LINKED ISSUE #55: clean up directory structure so tutorial things aren't called tests
remove /tests directory from top level

chloeann95:
closes #55


sbillinge (commented):
same here. Can we use the opportunity to write more tests?

sbillinge:
@sparky I am closing this.  I don't like that we put a binary zip file in here, so it makes more sense to me that we redo this PR from scratch.  We kind of want it that someone can clone the repo and run the examples following instructions in the docs or the README and not mess around with unzipping and so on, but that the tutorial examples are not huge files (which I don't think they are, but we should check).  Have a look and make a suggestion, make an issue of you have to.

PR #74, opened 2023-06-18
Change objective and reference naming

### LINKED ISSUE #37: MANT: change morph data naming
Current proposal:

`objective -> morphed` : data will be manipulated
`reference -> target`: data will stay unchanged.

The `obj-ref` language propagates across the repo (like `xref, yobj` etc), but the convention is rather clear and self-consistent as we always see `xref, yref` remain unchanged in the code.
 
The first order fix could be changing the language in `pdfmorph` function and app since they are user facing.

cc: @sbillinge @pavoljuhas

### LINKED ISSUE #46: Give file names instead of "reference" and "objective" in the plots

### LINKED ISSUE #55: clean up directory structure so tutorial things aren't called tests

Sparks29032:
Fixes #37. Fixes #46. Fixes #55.


diffpy/pdfmorph/pdfmorphapp.py
```diff
@@ -178,6 +178,24 @@ def custom_error(self, msg):
...
+        metavar="MLABEL",
+        dest="mlabel",
+        help="Set label for morphed data to MLABEL on plot. Ignored if using file names as labels.",
+    )
+    group.add_option(
+        '--tlabel',
+        metavar="TLABEL",
+        dest="tlabel",
+        help="Set label for target data to TLABEL on plot. Ignored if using file names as labels.",
+    )
```
Sparks29032, line 198: New option to set label names on plot.

diffpy/pdfmorph/morphs/morph.py
```diff
@@ -24,11 +24,11 @@
 
 
 class Morph(object):
-    '''Base class for implementing a morph on an objective given a reference.
+    '''Base class for implementing a morph given a target.
 
     Adapted from diffpy.pdfgetx to include two sets of arrays that get passed
-    through. In most cases, the objective is modified by a morph, but it is
-    acceptable for morph the reference as well, such as to change the range of
+    through. In most cases, only the morph is modified, but it is
```
sbillinge, line 30: Let's remove this sentence.  Changing the range of an array is not a morphing operation.

diffpy/pdfmorph/morphs/morphpdftordf.py
```diff
@@ -24,7 +24,7 @@
 class MorphXtalPDFtoRDF(Morph):
```
sbillinge, line 24: this is not a "morph" this is simply a transformation.  Please can we change this language?  I want to merge this so please make an issue to change the language (and name) on this.

Having said that, leave all the changes you have done and I will merge it,  but let's think a bit about the class name and the docstring/documentation.  I should probably understand exactly what is being done here before we finalize this, but stick it on an issue (and add it to the release milestone) so we loop back to it

diffpy/pdfmorph/morphs/morphrdftopdf.py
```diff
@@ -24,7 +24,7 @@
 class MorphXtalRDFtoPDF(Morph):
```
sbillinge, line 24: see above

diffpy/pdfmorph/pdfmorphapp.py
```diff
@@ -178,6 +178,24 @@ def custom_error(self, msg):
         dest="plot",
         help="Do not show the plot.",
     )
+    group.add_option(
+        '--usefilenames',
+        action="store_true",
+        dest="usefilenames",
+        help="Use the file names as labels on plot."
```
sbillinge, line 185: isn't this the default behavior?  So we don't need an option for this, just do this if one of the other options is not selected?

diffpy/pdfmorph/pdfmorphapp.py
sbillinge: looks good, see comment.   Please put different issues on different PRs

sbillinge (commented):
Changes look great, thanks!   

Please see quick inline comments and I can merge.

Please put less on each PR to make it easier to review and merge.  One thing per branch/pr ideally.  If something is  a quick cleaning thing you can add it to another one, but this should have been at least two PRs if not three.

Thanks so much

diffpy/pdfmorph/pdfmorphapp.py
Sparks29032: default behavior is to use "target" and "morph" as names

diffpy/pdfmorph/morphs/morph.py
Sparks29032: okie

diffpy/pdfmorph/morphs/morphpdftordf.py
Sparks29032: Yep, I'll open a new issue for this (and below)

Sparks29032:
> Changes look great, thanks!
> 
> Please see quick inline comments and I can merge.
> 
> Please put less on each PR to make it easier to review and merge. One thing per branch/pr ideally. If something is a quick cleaning thing you can add it to another one, but this should have been at least two PRs if not three.
> 
> Thanks so much

I will put less in each PR in the future, sorry for bundling them all together this time

diffpy/pdfmorph/morphs/morphpdftordf.py
Sparks29032: I am having trouble adding the issue to the milestone

diffpy/pdfmorph/pdfmorphapp.py
sbillinge: let's revisit this.  I will merge this and we can have a new quick issue after some discussion what is the best design

Issue #76, opened 2023-06-18
Change language in morphpdftordf and morphrdftopdf


Sparks29032:
Not morphs, just transformations. Change names of classes to reflect it.

sbillinge:
closed by #78

PR #78, opened 2023-06-20
Update terminology

Sparks29032:
Addresses #76. However, TransformPDFtoRDF and TransformRDFtoPDF still have 'morph' functions. Changing the names of these may be a bit difficult as we would have to change the terminology used by the MorphChain class.

We could change MorphChain -> ProcessChain and rename the function to 'process' rather than 'morph'. However, I prefer 'morph' since most classes (other than these two transforms) with the function are morphs.


sbillinge:
@Sparks29032 I  merging this.  Let's not worry about the mixing of morph and transformation for now.

Issue #79, opened 2023-06-22
Finalize morphshape and morphishape testing


Sparks29032:
Issue can be resolved after 0.0.1 release.
- Need to determine behavior at `psize` for `MorphISpheroid`. Do we (1) set to zero or (2) divide by characteristic function. Current test data has set to zero, but `_spheroidalCF` does not set the characteristic function to zero at the boundary.
- Add prolate spheroid test data.

sbillinge:
closed by #80

PR #80, opened 2023-06-22
Added test for MorphISpheroid, some metadata in test files

Sparks29032:
New issue raised involving testing MorphISpheroid see (#79).

Also, on `ni_qmax25_psize30.cgr`, G(r) is non-zero until 35, not 30. Should `psize` be changed to 35?


sbillinge:
> New issue raised involving testing MorphISpheroid see (#79).
> 
> Also, on `ni_qmax25_psize30.cgr`, G(r) is non-zero until 35, not 30. Should `psize` be changed to 35?

I think that is a yes on 30 vs 35.  Good catch.

sbillinge:
great @Sparks29032

PR #82, opened 2023-07-04
Add functionality for multiple morphs in one command

Sparks29032:
Fixes issue #84.

Added --sequence option for someone to morph a FILE multiple times against all files in a DIRECTORY as target. Plots the Rw values by default, and can save a table. Default is to sort files in DIRECTORY in alphabetical order.
image
image
A user can enable the --temperature option if all files in DIRECTORY end in "_###K.gr" or "_###K.cgr". This will sort files by ###, which should be the temperature (in Kelvin). Plots and tables look like the following if this option is enabled.
image
image

Fixes #77: default behavior is to set plot labels to file names. Making default just "morph" and "target" means multiple morphs in a morph sequence will have the same label.

Also fixes #47: checked the saving code enough to know it works.

New warnings appear during tests; see [issue](https://github.com/diffpy/diffpy.utils/issues/12) in diffpy.utils.

PR #83, opened 2023-07-19
Change libraries

Sparks29032:
Use Pytest instead of UnitTest, use importlib instead of pkg_resources (deprecated).


sbillinge:
> Seeing if we can move away from pkg_resouces. Use importlib.resources instead. Note importlib_resources does not work for some reason.

Thanks for this.   Have you googled around about this?  I had a conversation with Tom Caswell and he said that this change in python broke many things and all kinds of upstream packages have to make these changes.  I guess if we are using pkg_resources it would be good to fix this too, but maybe reach out to him if it is not discussed fully in your google searches.....

Sparks29032:
Changes merged to PR #82, branch no longer needed.

PR #89, opened 2024-05-02
Refactor src and switch to pyproject.toml

Sparks29032:


pyproject.toml
```diff
@@ -0,0 +1,51 @@
...
+        'Operating System :: Microsoft :: Windows',
+        'Operating System :: POSIX',
+        'Operating System :: Unix',
+        'Programming Language :: Python :: 3.10',
+        'Programming Language :: Python :: 3.11',
+        'Programming Language :: Python :: 3.12',
+	'Topic :: Scientific/Engineering :: Chemistry',
+        'Topic :: Scientific/Engineering :: Physics',
+]
+
```
sbillinge, line 33: we may need to add scripts or entry points I think.  Yucong ran into this with labpdfproc that it wouldn't run, then I added an entry point and resinstalled and it ran.

sbillinge (commented):
this looks great!  I left on comment inline.

Sparks29032:
Should be good now, added a check in `main.yml` to check running `pdfmorph` from command line

PR #91, opened 2024-05-17
Docstring Reformatting

Sparks29032:
Please merge before docs workflow is added. There will be errors in the `sphinx` build if not merged.

Some path fixes for packaged tutorial files and wholescale refactoring of docstrings into `numpy` format. Local documentation build looks good with this formatting.

Note that in-line documentation will not look good for all functions on all IDEs. Require a `Notes` section in order for the `numpy` documentation to "kick in" for some IDEs (e.g. `PyCharm`).


sbillinge:
Thanks for this work @Sparks29032 

I merged for all the good things.  @cadenmyers13 please spend a few minutes looking at the changes in the PR (you will find it as a closed PR which is a bit hard to find), to help you get used to some things like numpy docstrings and the way Andrew is doing things.

@Sparks29032 I merged this, thanks for the work.  A couple of thoughts.

1. I am trying to move to pathlib.Path() everywhere, so when you touch code that is using os.path it is a good opportunity to make that migration in general (though no need to go back and fix that here)
2. I will put a note in the other PR, but rather than fix the path in the workflow, it will save all kinds of hassle later if we just adopt a uniform directory structure.   I think `diffpy.utils` is more uniform with the other diffpy packages than pdfmorph, so I think it is better htat we move the directory structure here to be like there.  I will have Caden do that, maybe on a different PR.

PR #96, opened 2024-05-22
Ensure evaluation

Sparks29032:
Need `${{}}` otherwise the inner conditional will not be evaluated (without evaluation, the `if` statement evaluates the string literal which will always return true).

Also update `peaceiris` version and include `release` as a trigger for deploying documentation.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -42,7 +42,7 @@ jobs:
         run: make -C doc/manual html
 
       - name: Deploy
-        if: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'release') }}
+        if: ${{ (github.event_name == 'push' && github.ref == 'refs/heads/main') || github.event_name == 'release' }}
         uses: peaceiris/actions-gh-pages@v4
         with:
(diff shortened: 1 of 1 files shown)


sbillinge:
thanks @Sparks29032

sbillinge:
@cadenmyers13 can we put these same things into diffpy.utils and also diffpy.labpdfproc?  In principle we would like to add them to all the diffpy projects!   Another project is to create a cookiecutter for group releases and that could have these in them so all future repos look this way by default.

cadenmyers13:
> @cadenmyers13 can we put these same things into diffpy.utils and also diffpy.labpdfproc? In principle we would like to add them to all the diffpy projects! Another project is to create a cookiecutter for group releases and that could have these in them so all future repos look this way by default.

Yep. ill take care of it!

cadenmyers13:
> > @cadenmyers13 can we put these same things into diffpy.utils and also diffpy.labpdfproc? In principle we would like to add them to all the diffpy projects! Another project is to create a cookiecutter for group releases and that could have these in them so all future repos look this way by default.
> 
> Yep. ill take care of it!

Looks like Andrew took care of it for diffpy.utils, ill put it in diffpy.labpdfproc

PR #104, opened 2024-05-29
Edits made by running pre-commit

cadenmyers13:
Made some manual edits as well to prevent undesired removal of whitespace.


sbillinge (commented):
this looks good.  It seems I should have run black on the docs but I didn't,  but np, it is done here.

Please see the inline comments.

I think I would like to merge this before you start all the manual flake8 work just to keep the PR more under control. Is that ok?  So just fix the comments and then I will merge and then you can start fixing the flake8 stuff on a separate PR.  We can also go bit by bit on that too since we are doing a bunch of remedial syntax cleaning but everything is still passing non linting tests.

PR #110, opened 2024-05-30
Fix final errors in Flake8

cadenmyers13:
Just a few errors about spacing. Pre-commit is passing now!


sbillinge:
These space after colons are in formatting statements so the spaces are
likely to change behavior.  Could you maybe check this by running the code
with and without that change and make sure the output is the same.  If not
we may have discovered why this check was switched off in flake8 before...

src/diffpy/pdfmorph/pdfmorphapp.py
```diff
@@ -104,7 +104,7 @@ def custom_error(self, msg):
         help=f"""Changes usage to \'{prog_short} [options] FILE DIRECTORY\'. FILE
  will be morphed with each file in DIRECTORY as target.
  Files in DIRECTORY are sorted by alphabetical order unless a field is
- specified by --sort-by.""",
+specified by --sort-by.""",
```
Sparks29032, line 107: We need these spacings for displaying options. Similar for the other.

PR #114, opened 2024-06-03
Ignore errors in .flake8

cadenmyers13:
Ignore formatting errors E231, E271, E272 regarding space in docstrings and around ':'

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -9,4 +9,5 @@ exclude =
     doc/manual/source/conf.py
 max-line-length = 115
 # Ignore some style 'errors' produced while formatting by 'black'
-ignore = W503
+# E231, E271, and E272 are ignored for formatting purposes regarding space around ':' and docstrings
+ignore = W503, E231, E271, E272
--- a/.flake8
+++ b/.flake8
@@ -9,5 +9,8 @@ exclude =
     doc/manual/source/conf.py
 max-line-length = 115
 # Ignore some style 'errors' produced while formatting by 'black'
-# E231, E271, and E272 are ignored for formatting purposes regarding space around ':' and docstrings
+# W503: line break error in morphshape.py
+# E231: Missing whitespace after ':'. Ignored for formatting in pdfmorph_io.py (i.e. lines 64, 213, 236)
+# E271: Multiple spaces before keyword. Ignored for docstring formatting in pdfmorphapp.py
+# E272: Multiple spaces after keyword. Ignored for docstring formatting in pdfmorphapp.py
 ignore = W503, E231, E271, E272


.flake8
```diff
@@ -9,4 +9,5 @@ exclude =
     doc/manual/source/conf.py
 max-line-length = 115
 # Ignore some style 'errors' produced while formatting by 'black'
-ignore = W503
+# E231, E271, and E272 are ignored for formatting purposes regarding space around ':' and docstrings
```
sbillinge, line 12: please put all the details.  Which error is precisely for what and why we ignore it.

sbillinge (commented):
pls see comment.

.flake8
```diff
@@ -9,5 +9,8 @@ exclude =
     doc/manual/source/conf.py
 max-line-length = 115
 # Ignore some style 'errors' produced while formatting by 'black'
-# E231, E271, and E272 are ignored for formatting purposes regarding space around ':' and docstrings
+# W503: line break error in morphshape.py
+# E231: Missing whitespace after ':'. Ignored for formatting in pdfmorph_io.py (i.e. lines 64, 213, 236)
```
sbillinge, line 13: This is not quite what I am after.  If a test is erroring we can fix it in two ways.  We can change the test or we can change the code so the test passes.  The ONLY reason to change the test is to make the code better so you can't use something along the lines of "ignored to make the test pass in pdfmorph" as a justification for changing the test. You are only allowed reasons such as "this changes formatting in f-strings that results in formatting errors in the resulting print statements."  In this latter case, it is good to understand actually what those errors look like.  Since we are learning this for the first time here, it is worth actually doing the test.  i.e., running the code and looking at the print statement with and without the space in there to see actually what happens.....

The same is true for the other skips.  Do the test, see what happens and then summarize it in the comment.   When we are satisfied we will adopt this across all the repos.

sbillinge (commented):
pls see inline

.flake8
sbillinge: @cadenmyers13 @Sparks29032 I am still a bit nervous about making all these exceptions.   I did some digging and found this....it seems like it may be a better fix for the E231:
https://stackoverflow.com/questions/77401175/how-to-make-flake8-ignore-syntax-within-strings

We do want to skip W503 which is deprecated.

Can we find the fixes for E271 and E272 and not skip them?

sbillinge (commented):
please see inline

.flake8
cadenmyers13: @sbillinge I'll look into E271 and E272. I tried what was recommended in the link you sent and I still got the same error. I'll continue searching to see if I can find anything else that might fix this.

Sparks29032:
We can probably close this?

PR #116, opened 2024-06-07
Add tutorial for morph sequence and shape

Sparks29032:
Add tutorial into `quickstart.rst` detailing how to use the multiple morphs function (perform morphs to compare all files in a directory against some target file) and morphshape.

Update `README.rst` instructions so they no longer reference the `diffpy` channel.

Remove `__date__` from `version.py` as the assumption made is wrong and we do not use  `__date__` in the code (can be made into a separate PR upon request).


src/diffpy/pdfmorph/refine.py
```diff
@@ -114,14 +114,14 @@ def refine(self, *args, **kw):
...
-            raise ValueError(mesg)
+        sol, cov_sol, infodict, emesg, ier = leastsq(self.residual, initial, full_output=1)
+        fvec = infodict["fvec"]
+        if ier not in (1, 2, 3, 4):
+            emesg
+            raise ValueError(emesg)
 
         # Place the fit parameters in config
-        vals = out[0]
+        vals = sol
```
Sparks29032, line 124: For better readability in the future.

Sparks29032:
Thanks for catching the typos!

PR #121, opened 2024-06-12
Fix behavior for multiple morphs

Sparks29032:
Originally programmed to error if there are non-PDF files (or directories), better not to error when directory is there since user may save their morphed PDFs in the same directory within a subdirectory.

Also rename outputted `.csv` to `.txt` as it is definitely not a `.csv`. The format can be read by `loadData`, so I think it is fine not to change it to an explicit `.csv`.

Also reorder the `--help` order so options involving morphs available and plotting are above options for doing multiple morphs.


src/diffpy/pdfmorph/pdfmorphapp.py
```diff
@@ -53,7 +53,12 @@ def custom_error(self, msg):
                 "Use --help for help.",
             ]
         ),
-        epilog="Please report bugs to diffpy-dev@googlegroups.com.",
+        epilog="\n".join(
+            [
+                "For more information, see the PDFmorph manual."
+                "Please report bugs to diffpy-dev@googlegroups.com.",
```
sbillinge, line 59: we don't use this any more so it would be good to remove diffpy-dev from everywhere.  Let's ask them to report questions and bugs to diffpy-users.

Can we also do a global search and make this change everywhere in all the projects.

sbillinge (commented):
please see inline comment.

src/diffpy/pdfmorph/pdfmorphapp.py
Sparks29032: Changed everywhere in `PDFmorph`. I put in an [issue](https://github.com/bg-mission-control/operation-release-everything/issues/6) to `operation-release-everything` about the remaining instances of `diffpy-dev` that need to be updated.

PR #131, opened 2024-09-12
Manual Math Error Fix

Sparks29032:
Fixing math error in the manual. Change requirements from `matplotlib-base` to `matplotlib`.

Using `matplotlib-base` breaks the command `pip install -r https://raw.githubusercontent.com/diffpy/diffpy.pdfmorph/main/requirements/run.txt` in our installation instructions in the README.rst.


sbillinge (commented):
Is there no way to make the matplotlib-base work?  I would only like to require full matplotlib if we need full matplotlib (i.e., for plotting with a gui)

Sparks29032:
There's no `matplotlib-base` for `PyPi`. On our `conda` recipe, we only use `matplotlib-base` in the feedstock.

Unless we change the `pip` install instructions `pip install -r https://raw.githubusercontent.com/diffpy/diffpy.pdfmorph/main/requirements/run.txt`, we need `matplotlib` in our `run.txt.`. Changing it in `run.txt` won't affect our `conda` build though, that will still use the bare bones one.

sbillinge:
I reached to Tom Caswell for advice on this.   We don't want to ship the full matplotlib if we can avoid it to reduce bloat.  He may have a recommendation.

sbillinge:
This seems to be failing tests.  It is not installing then importing the stylesheet correctly.  Please can you take a look.   

Actually we need a way to delete the package `bg_mpl_stylesheets`, or better, have it resolve to `bg-mpl-stylesheets`, but either way, we kind of need pdfmorph to install and import correctly.

Sparks29032:
@sbillinge Fixed in #132, this PR only changes the Manual (files changed are just the `texinfo`, and `pdf`).

PR #132, opened 2024-10-02
Cookie Cutter + PyPi fix

Sparks29032:
Cookiecutter, standard stuff.

Also added contents to `pip.txt`. This is the same as `run.txt` except `matplotlib-base` is changed to `matplotlib`. This is because `matplotlib-base` is not available on `PyPi`, and cannot be `pip` or `pipx` (Tom's suggestion) installed. This does not bloat our `conda-forge` release, which uses `run.txt`, but it ensures anything building or installing with `pip` pulls the correct package.

Below is the result without this fix:
image
image


Sparks29032:
Got it, I'll consolidate pre-commit next time.

PR #145, opened 2024-10-29
Adding multiple morphs for single target

Sparks29032:
Originally only have multiple targets for a single file to be morphed to. See news item.


Sparks29032:
Force pushes may occur as the pre-commit is not set up within the 202420sacla_expt directory.

Sparks29032:
Prior to merging: must recompile manual pdf version and put in both locations.

Sparks29032:
Good to merge.

PR #153, opened 2024-11-10
Fix job names in `check-news-item.yml`, `tests-on-pr.yml`

bobleesj:
Also testing whether codecov comment here works.


bobleesj:
https://github.com/diffpy/diffpy.pdfmorph/pull/152#issuecomment-2466677407

In fact codecov failing was a glitch...

PR #155, opened 2024-11-12
Add shifting morph

Sparks29032:
There was already a module for shifting morphs, but it has now been integrated into the refinement stack. Tests have been added/modified for this new morph.

### DIFF
--- a/src/diffpy/pdfmorph/morphs/morphshift.py
+++ b/src/diffpy/pdfmorph/morphs/morphshift.py
@@ -47,9 +47,13 @@ class MorphShift(Morph):
 
     def morph(self, x_morph, y_morph, x_target, y_target):
         """Apply the shifts."""
-        if self.hshift is None:
+        try:
+            self.hshift
+        except AttributeError:
             self.hshift = 0
-        if self.vshift is None:
+        try:
+            self.vshift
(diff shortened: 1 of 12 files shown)


src/diffpy/pdfmorph/morphs/morphshift.py
```diff
@@ -47,10 +47,19 @@ class MorphShift(Morph):
...
         """Apply the shifts."""
+        try:
+            hshift = self.hshift
+        except AttributeError:
+            hshift = 0
+        try:
+            vshift = self.vshift
+        except AttributeError:
+            vshift = 0
+
```
Sparks29032, line 58: This code ensures that the vertical/horizontal shifts can be done independent of each other. We can also create two classes: MorphHShift and MorphVShift, but I feel that is overkill.

Sparks29032:
Ready to merge after review.

sbillinge:
> CodeCov is being silly... the untested line missing coverage is the
> 
> ```
> if __name__ == "__main__":
>     TestMorphShift()
> ```
> 
> line at the end of the test module.

I think there is a way to have it avoid this, but I don't remember how.  But for now, I will just merge it.

PR #157, opened 2024-12-08
Fix for multiple morphs

Sparks29032:
Used to break for multiple subdirectories within the many-morphs directories.


sbillinge (commented):
this probably needs tests, or is it being tested but the tests didn't need to be changed.  If so, ere they failing before?  If not they need to be extended....

Sparks29032:
Yes, tests needed. I think adding 2 directories with blank files to testsequence is sufficient if the tests still pass as there is no specific output given when directories are parsed out.

Sparks29032:
Tests pass. Good to merge?

sbillinge (commented):
please see inline

Sparks29032:
Closing for fresh bugfix branch.

PR #168, opened 2025-03-04
docs: remove texinfo

### LINKED ISSUE #167: docs: remove texinfo

ycexiao:
closes #167


ycexiao:
close #167

sbillinge:
> close #167

this should be "closes #167" and to work it needs to be in the top-most comment box, right below the title.  You can add it there by editing that box.

README.rst
```diff
@@ -137,9 +137,6 @@ You may consult our `online documentation <https://diffpy.github.io/diffpy.pdfmo
 USING PDFmorph
 ------------------------------------------------------------------------
 
-For detailed instructions and full tutorial, consult the user manual
```
sbillinge, line 140: I think we still would like a link to the manual.  So this shouldn't be deleted but modified to point to where the new manual will be.

If we don't quite know that yet, please create an issue to fix that later, then we can resolve this conversation.  But if possible, fix it.

doc/source/quickstart.rst
```diff
@@ -4,8 +4,7 @@ PDFmorph Tutorial
 #################
 
 Welcome! This will be a quick tutorial to accquaint users with PDFmorph
-and some of what it can do. For a more detailed tutorial, check out
-our :download:`user manual <../manual/pdfmorph.pdf>`.
+and some of what it can do.
```
sbillinge, line 7: see above, I think we need a link to the new manual here.

sbillinge (commented):
Thanks @ycexiao for this.  To make sure everything is working and make it easier to review, it might make sense to put screenshots of the previous manual and how it looks in the new form.  Then we can be sure that everything is there and nothing is lost.  If the manual is long, we could just do the top of each section or something like that?

ycexiao:
New, section beginning
image
New, insert image
image

Old, section beginning
image
Old, insert image
image

ycexiao:
@sbillinge Please review this. Old pdfmorph.pdf is reserved in the `source` directory. Several screenshots about the comparison are shown above.

doc/source/index.rst
```diff
@@ -39,7 +39,8 @@ the plotted PDFs.
 Finally, we note that though ``PDFmorph`` should work on other spectra
 that are not PDFs, it has not been extensively tested beyond the PDF.
 
-To get started, please visit the :ref:`quick_start`.
+To get started, please download our :download:`user manual  <pdfmorph.pdf>`
```
sbillinge, line 42: I think the pdf version won't exist any more shortly, right?  Just the version in the online docs.  Isn't this the goal of your project?

doc/source/quickstart.rst
```diff
@@ -4,7 +4,8 @@ PDFmorph Tutorial
 #################
 
 Welcome! This will be a quick tutorial to accquaint users with PDFmorph
-and some of what it can do.
+and some of what it can do. For a more detailed tutorial, check out
+our :download:`user manual <pdfmorph.pdf>`.
```
sbillinge, line 8: please see above.  no more texinfo so no more PDF version of the manual?   Do we need to talk again to make sure our goals are aligned?

sbillinge (commented):
please see my latest inline comments

doc/source/index.rst
ycexiao: Yes there will only be the online version. I think I misunderstood your previous comment, which said there should be a link to the manual. So I recovered the older version of pdf.

Now I am a little unsure, do we need the pdf manual built by rst? Because the information already exists and is better rendered in the online documents. Or just leave the link to the online documents.

sbillinge: As guidance, maybe remember your goal and your target audience.  It is a new user of PDFmorph.  They have read the introduction and are excited to use it.  What to do next?  I think that person might like to given some possible "next steps" to get started.  So if they see something like:  
`To get started, please visit the :ref:`quick_start`.` I think it could be helpful, if when they click it and go there, there are nice and clear steps about how to get started.  Did you test that "flow" as if you were a new user and see if you think it is easy to follow and makes sense?  That is the task here that we are looking for.

Also, because we want to remove the PDF/texinfo manual, we want to double check that there is nothing in there that is NOT in the rst version of the manual that we might lose and don't want to lose.  That can be on another PR.

Have you used PDFmorph?  Do you know what it is for and how it works?  This is probably important or you won't be able to do this job well.

ycexiao: Thanks for instructions. I see that the job here is more than just removing the files and replacing with the new ones. I will get into the repo itself and gain more insights about learning and using it through the old and new manuals.

sbillinge:
@ycexiao I wonder if we want to create a new clean PR for this work now that it is better understood.  What do you think? 

On another note, I see a lot of image files being deleted.  Is this because they exist in another place and are replaced, or are these images that are not used in any manual?

ycexiao:
> @ycexiao I wonder if we want to create a new clean PR for this work now that it is better understood. What do you think?
> 
> On another note, I see a lot of image files being deleted. Is this because they exist in another place and are replaced, or are these images that are not used in any manual?

Sure. And I can follow the correct concept working on this issue in the new PR, too.

About the images.The situation is that the last several sections are removed. According to Andrew, these sections are focused on the theory derivation, which are little off topic and they can also be found in cited papers. I think in the new manuals we should leave these complex theories aside and just provide clear citation to these papers. Also considering the time need to configure the .rst to properly render math equations, these sections are removed and images are deleted.

sbillinge:
> > @ycexiao I wonder if we want to create a new clean PR for this work now that it is better understood. What do you think?
> > On another note, I see a lot of image files being deleted. Is this because they exist in another place and are replaced, or are these images that are not used in any manual?
> 
> Sure. And I can follow the correct concept working on this issue in the new PR, too.
> 
> About the images.The situation is that the last several sections are removed. According to Andrew, these sections are focused on the theory derivation, which are little off topic and they can also be found in cited papers. I think in the new manuals we should leave these complex theories aside and just provide clear citation to these papers. Also considering the time need to configure the .rst to properly render math equations, these sections are removed and images are deleted.

sounds good.  Please make sure to add text explaining where people can find that information with citations to the papers.

PR #171, opened 2025-03-18
docs: add temperature field to tutorial data

### LINKED ISSUE #169: docs: additional tutorial data for multiple morph misses `temperature` field
Both the old and new manual uses
```
pdfmorph SrFe2As2_150K.gr . --multiple-targets --sort-by=temperature
```
to demonstrate how to sorting Rw by assigned field.

However, by running this I get
```
pdfmorph: error: The requested field is missing from a PDF file header.
```

According to the old manual section 3.2.1, page 10, the expected .gr should start with

Image

But the data now in tutorial/additionalData/morphsequence/ , it start with

Image

Temperature field is missed.

Also in the manual the corresponding directory is referred as `morphMultiple`, it  seems that it is renamed as `morphsequence`. I can fix that in the other PR.

Sorting morphs by temperature is a good example. The only problem is that the data we are using doesn't have the `temperature` field now. I can add them manually, or someone could  provide me the correct version of these data.

ycexiao:
close #169


ycexiao:
@sbillinge It's ready for review.

I think the data should be update as soon as possible and I will open another issue to suggest a feature that can take file names as an input to extract meta data.

news/data_temperature.rst
```diff
@@ -0,0 +1,23 @@
+**Added:**
+
+* add temperature field to tutorial/additionalData.
```
sbillinge, line 3: this should be in fixed.  This is not an added functionality.

sbillinge (commented):
please see inline

ycexiao:
codecov will go wrong  because the upstream name changed. Redo the PR.

PR #172, opened 2025-03-19
Refactor PDFmorph to diffpy.morph

Sparks29032:
All tests pass on local. Links need to be fixed by refactoring this repository from `diffpy.pdfmorph` to `diffpy.morph`.


Sparks29032:
@sbillinge Should only require renaming this repository before we can merge.

PR #174, opened 2025-03-20
fix: add temperature field to tutorial data

### LINKED ISSUE #169: docs: additional tutorial data for multiple morph misses `temperature` field

ycexiao:
close #169


ycexiao:
@sbillinge It's ready form review.

PR #178, opened 2025-03-28
test: created a test for squeeze morph using sine wave. Created funct…

Luiskitsu:
I have created a test, test_morphsqueeze.py, and a function, morpsqueeze.py, for squeeze morph. This has been now applied to a sine wave and the test run successfully. I have written the test and the function based on the other morphs. Next step is to change the function to use polynomials. I am not sure how do you go from this new morph class to update the main code? Is this done automatically?

### DIFF
--- a/news/morphsqueeze.rst
+++ b/news/morphsqueeze.rst
@@ -5,7 +5,9 @@
 **Changed:**
 
 * Added cubic and quadratic terms, as well as
-* squeeze_1 and squeeze_2 parameters
+* squeeze_1 and squeeze_2 parameters.
+* I was not keeping the y-axis constant and
+* corrected for it
 
 **Deprecated:**
 
@@ -17,7 +19,7 @@
 
 **Fixed:**
 
-* <news item>
+* I was not keeping the y-values constant, fixed now
 
 **Security:**
 
--- a/news/morphsqueeze.rst
+++ b/news/morphsqueeze.rst
@@ -4,7 +4,8 @@
 
 **Changed:**
 
-* Nothing changed
+* Added cubic and quadratic terms, as well as
+* squeeze_1 and squeeze_2 parameters
 
 **Deprecated:**
 
--- a/news/morphsqueeze.rst
+++ b/news/morphsqueeze.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* Squeeze morph, test with sine wave
+
+**Changed:**
+
+* Nothing changed
+
+**Deprecated:**
+
+* <news item>
+
+**Removed:**
+
+* <news item>
+
+**Fixed:**
+
+* <news item>
+
+**Security:**
+
+* <news item>
--- a/tests/test_morphsqueeze.py
+++ b/tests/test_morphsqueeze.py
@@ -1,0 +1,56 @@
+import numpy as np
+import pytest
+
+from diffpy.morph.morphs.morphsqueeze import MorphSqueeze
+
+
+class TestMorphSqueeze:
+    @pytest.fixture
+    def setup(self):
+        # Create a sine-wave for testing
+        self.x_morph = np.linspace(0, 2 * np.pi, 1000)
+        self.y_morph = np.sin(self.x_morph)
+        self.x_target = self.x_morph.copy()
+        self.y_target = self.y_morph.copy()
+        return
+
+    def test_no_squeeze(self, setup):
+        """When squeeze is zero, the input should be unchanged."""
+        morph = MorphSqueeze()
+        morph.squeeze = 0.0
+
+        x_morph, y_morph, x_target, y_target = morph(
+            self.x_morph, self.y_morph, self.x_target, self.y_target
+        )
+
+        # Verify that the morph output matches the original input
+        assert np.allclose(x_morph, self.x_morph)
+        assert np.allclose(y_morph, self.y_morph)
+        # And the target arrays remain unchanged
+        assert np.allclose(x_target, self.x_target)
+        assert np.allclose(y_target, self.y_target)
+
+    def test_morph_with_squeeze(self, setup):
+        """Test that with a non-zero squeeze,
+        x_morph is transformed non-linearly."""
+        morph = MorphSqueeze()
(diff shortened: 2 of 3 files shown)


Luiskitsu:
I forgot to write the News section!

Luiskitsu:
Just updated the code to do the cubic and quadratic non-linear squeeze to the x-axis. I have updated the code to have squeeze_1 and squeeze_2. I have also tested the code (hopefully correctly?)

Luiskitsu:
The function now applies a non-linear transformation to the x-axis (x + squeeze_1*x^2 + squeeze_2*x^3). The results are the same y-axis values in a non-linearly transformed x-axis. Would be good to plot the results in the test to check how it looks.

Luiskitsu:
Will need to interpolate if target data has different x-axis values than the data we want to morph. For our current experiment, the target data goes up to 22 1/A and the data we want to morph goes up to ~11 1/A. We want to interpolate the target data to the morph data.

news/morphsqueeze.rst
```diff
@@ -0,0 +1,26 @@
+**Added:**
+
+* Squeeze morph, test with sine wave
+
+**Changed:**
+
+* Added cubic and quadratic terms, as well as
```
sbillinge, line 7: none of these are needed.  We just need one news item per PR basically, not a history of everything that happened on the way.  So what is needed is, under Added, something like "Squeeze morph that carries out a non-linear stretch to the independent variable axis."

tests/test_morphsqueeze.py
```diff
@@ -0,0 +1,65 @@
+import numpy as np
+import pytest
+
+from diffpy.morph.morphs.morphsqueeze import MorphSqueeze
+
+
+class TestMorphSqueeze:
```
sbillinge, line 7: What is your reason for making a class here?  We usually just write simpler functions but annotate them to show the intended behavior.  Please see `diffpy.utils` for examples

tests/test_morphsqueeze.py
```diff
@@ -0,0 +1,65 @@
...
+    @pytest.fixture
+    def setup(self):
```
sbillinge, line 9: I don't think this needs to be a fixture.  If we are going to reuse it in multiple places we can make it a fixture, but then put it in conftest.py

tests/test_morphsqueeze.py
```diff
@@ -0,0 +1,65 @@
...
+        morph = MorphSqueeze()
+        morph.squeeze_1 = 0.0
+        morph.squeeze_2 = 0.0
+
+        x_morph, y_morph, x_target, y_target = morph(
+            self.x_morph, self.y_morph, self.x_target, self.y_target
+        )
+
+        # Verify that the morph output matches the original input
+        assert np.allclose(x_morph, self.x_morph)
```
sbillinge, line 28: this is easier to follow using the `expected` and `actual` syntax

tests/test_morphsqueeze.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+        expected_y = np.sin(self.x_morph)
+
+        res = sum(np.fabs(expected_y - y_new))
+        res2 = sum(np.fabs(expected_x - x_new))
+        assert res < 0.1
+        assert res2 < 0.1
+
+
+if __name__ == "__main__":
```
sbillinge, line 64: this not needed

sbillinge (commented):
great start.  Please see my inline comments.

My main comment is that we may be going too fast here.  I don't see a discussion of what we want to accomplish here and then clear tests of that.  I am not sure if you are using ChatGPT but it is looking a bit like what would come from there.   We don't really want to defer architectural decisions to chatgpt, though it can be useful as a first cut to specific code blocks.

We want to make sure to be consistent across morphs.    There is always a choice between keeping the x-array constant and varying the y-array or vice versa.   the stretch morph seems to be transforming on the y-array, so  I think it may be a bad idea to transform the x-array here. 

to be confident it is working I would go all the way of creating an input that is squeezed but on a uniform grid, because that is the data that we will be getting.  I.e., I imagine that we will do an experiment that will give a uniform Q-grid, but because of a weird calibration boo boo, the signal has been squeezed.   The transform we want to do is to unsqueeze it so we can compare it to synchrotron data that has no such calibration error.  But in general we may want to operate on those arrays (like take a difference etc.) so we would like them to be on the same grid as each other.   

Just some thoughts.

If this is what we want, let's devise a test that does this.  Then the code.

Luiskitsu:
@sbillinge thanks for the feedback. I was using test_morphstretch.py and morphstretch.py as my basis, that is why I made a class for TestMorphSqueeze, made a fixture, etc. I was changing the x-axis instead of the y-axis as I thought was more intuitive but makes sense to be consistent with the other morphs. I will look at diffpy.utils for examples, I think this will really help me here as I am not used to the syntax or the tests.
Do you mean I should first make a test without calling any morph function? I will follow all your comments and make a new test as you mention.

tests/test_morphsqueeze.py
Luiskitsu: I was following test_morphstretch.py. Will change this

Luiskitsu:
@sbillinge  Should I create a new PR where I can add a test and we can go step by step? Or continue with this PR? I will generate a test where we have an expected uniform signal, such as a sine wave, our input data will be a squeezed version of the sine wave, and then I will unsqueeze it.

sbillinge:
> @sbillinge Should I create a new PR where I can add a test and we can go step by step? Or continue with this PR? I will generate a test where we have an expected uniform signal, such as a sine wave, our input data will be a squeezed version of the sine wave, and then I will unsqueeze it.

yes, this is a good approach.

sbillinge:
for this first test, I would create the squeezed target and then plot it on top of the unsqueezed one and paste that in the comments on the PR so we have a visual about what it looks like.  We won't do any plotting as part of the test, but it is part of the workflow of building a good test that we are confident is testing what we want to test.

sbillinge:
closing.  replaced by #182

PR #182, opened 2025-04-11
Squeeze morph: Adding UCs tests

Luiskitsu:
I have written different UCs tests from issue 181 for the squeeze morph. I am keeping the morphed x-axis the same as the unmorphed as we discussed. I have left the morph function empty. My inputs are a uniform target, and a squeezed morph, then the expected morph is the unsqueezed morph.

### DIFF
--- a/src/diffpy/morph/morphs/morphsqueeze.py
+++ b/src/diffpy/morph/morphs/morphsqueeze.py
@@ -16,6 +16,10 @@ class MorphSqueeze(Morph):
     xoutlabel = LABEL_RA
     youtlabel = LABEL_GR
     parnames = ["squeeze"]
+    # extrap_index_low: last index before interpolation region
+    # extrap_index_high: first index after interpolation region
+    extrap_index_low = None
+    extrap_index_high = None
 
     def morph(self, x_morph, y_morph, x_target, y_target):
         """Squeeze the morph function.
@@ -32,13 +36,10 @@ class MorphSqueeze(Morph):
 
         Returns
         -------
-            A tuple (x_morph_out, y_morph_out, x_target_out, y_target_out,
-            min_index, max_index) where the target values remain the same and
-            the morph data is shifted according to the squeeze. The morphed
-            data is returned on the same grid as the unmorphed data.
-            The min_index and max_index are the last index before the
-            interpolated region and the first index after the interpolated
-            region, respectively. If there is no extrapolation it returns None.
+            A tuple (x_morph_out, y_morph_out, x_target_out, y_target_out)
+            where the target values remain the same and the morph data is
+            shifted according to the squeeze. The morphed data is returned on
+            the same grid as the unmorphed data.
 
         Example
         -------
@@ -66,8 +67,8 @@ class MorphSqueeze(Morph):
         self.y_morph_out = CubicSpline(x_squeezed, self.y_morph_in)(
             self.x_morph_in
         )
-        left_extrap = np.where(self.x_morph_in < x_squeezed[0])[0]
-        right_extrap = np.where(self.x_morph_in > x_squeezed[-1])[0]
-        min_index = left_extrap[-1] if left_extrap.size else None
-        max_index = right_extrap[0] if right_extrap.size else None
-        return self.xyallout + (min_index, max_index)
+        low_extrap = np.where(self.x_morph_in < x_squeezed[0])[0]
+        high_extrap = np.where(self.x_morph_in > x_squeezed[-1])[0]
+        self.extrap_index_low = low_extrap[-1] if low_extrap.size else None
+        self.extrap_index_high = high_extrap[0] if high_extrap.size else None
+        return self.xyallout
--- a/src/diffpy/morph/morphs/morphsqueeze.py
+++ b/src/diffpy/morph/morphs/morphsqueeze.py
@@ -4,12 +4,8 @@ from diffpy.morph.morphs.morph import LABEL_GR, LABEL_RA, Morph
 class MorphSqueeze(Morph):
     """Squeeze the morph function.
 
-    This applies a polynomial to squeeze the morph non-linearly. The resulting
-    squeezed morph is interpolated to the (trimmed) target grid.
-    Only the overlapping region between the squeezed morph and the target
-    grid is used. The target is trimmed (or not) accordingly, and the final
-    outputs (morph and target) are returned on the same grid, defined by this
-    trimmed target range.
+    This applies a polynomial to squeeze the morph non-linearly. The morphed
+    data is returned on the same grid as the unmorphed data.
 
     Configuration Variables
     -----------------------
@@ -28,9 +24,5 @@ class MorphSqueeze(Morph):
 
     def morph(self, x_morph, y_morph, x_target, y_target):
         Morph.morph(self, x_morph, y_morph, x_target, y_target)
-        if self.squeeze is None:
-            self.x_morph_out = self.x_morph_in
-            self.y_morph_out = self.y_morph_in
-            return self.xyallout
 
         return self.xyallout
--- a/src/diffpy/morph/morphs/morphsqueeze.py
+++ b/src/diffpy/morph/morphs/morphsqueeze.py
@@ -2,37 +2,8 @@ from diffpy.morph.morphs.morph import LABEL_GR, LABEL_RA, Morph
 
 
 class MorphSqueeze(Morph):
-    """Squeeze the morph function.
-
-    This applies a polynomial to squeeze the morph non-linearly. The morphed
-    data is returned on the same grid as the unmorphed data.
-
-    Configuration Variables
-    -----------------------
-    squeeze : list
-        The polynomial coefficients [a0, a1, ..., an] for the squeeze function
-        where the polynomial would be of the form a0 + a1*x + a2*x^2 and so
-        on.  The order of the polynomial is determined by the length of the
-        list.
-
-    Example
-    -------
-    >>> import numpy as np
-    >>> from numpy.polynomial import Polynomial
-    >>> from diffpy.morph.morphs.morphsqueeze import MorphSqueeze
-
-    >>> x_target = np.linspace(0, 10, 101)
-    >>> y_target = np.sin(x_target)
-    >>> x_morph = np.linspace(0, 10, 101)
-    >>> squeeze_coeff = [0.1, -0.01, 0.005]
-    >>> poly = Polynomial(squeeze_coeff)
-    >>> y_morph = np.sin(x_morph + poly(x_morph))
-
-    >>> morph = MorphSqueeze()
-    >>> morph.squeeze = squeeze_coeff
-    >>> x_morph_out, y_morph_out, x_target_out, y_target_out = morph(
-    ...     x_morph, y_morph, x_target, y_target)
-    """
+    """Apply a polynomial to squeeze the morph function. The morphed
+    data is returned on the same grid as the unmorphed data."""
 
     # Define input output types
     summary = "Squeeze morph by polynomial shift"
@@ -43,7 +14,44 @@ class MorphSqueeze(Morph):
     parnames = ["squeeze"]
 
     def morph(self, x_morph, y_morph, x_target, y_target):
-        """Apply a polynomial to squeeze the morph function"""
+        """Squeeze the morph function.
+
+        This applies a polynomial to squeeze the morph non-linearly.
+
+        Configuration Variables
+        -----------------------
+        squeeze : list
+            The polynomial coefficients [a0, a1, ..., an] for the squeeze
+            function where the polynomial would be of the form
+            a0 + a1*x + a2*x^2 and so on.  The order of the polynomial is
+            determined by the length of the list.
+
+        Returns
+        -------
+            A tuple (x_morph_out, y_morph_out, x_target_out, y_target_out)
+            where the target values remain the same and the morph data
+            is shifted according to the squeeze. The morphed data is
+            returned on the same grid as the unmorphed data.
+
+        Example
+        -------
+        Import the squeeze morph function:
+        >>> from diffpy.morph.morphs.morphsqueeze import MorphSqueeze
+        Provide initial guess for squeezing coefficients:
+        >>> squeeze_coeff = [0.1, -0.01, 0.005]
+        Run the squeeze morph given input morph array (x_morph, y_morph)
+        and target array (x_target, y_target):
+        >>> morph = MorphSqueeze()
+        >>> morph.squeeze = squeeze_coeff
+        >>> x_morph_out, y_morph_out, x_target_out, y_target_out = morph(
+        ...     x_morph, y_morph, x_target, y_target)
+        To access parameters from the morph instance:
+        >>> x_morph_in = morph.x_morph_in
+        >>> y_morph_in = morph.y_morph_in
+        >>> x_target_in = morph.x_target_in
+        >>> y_target_in = morph.y_target_in
+        >>> squeeze_coeff_out = morph.squeeze
+        """
         Morph.morph(self, x_morph, y_morph, x_target, y_target)
 
         return self.xyallout
--- a/src/diffpy/morph/morphs/morphsqueeze.py
+++ b/src/diffpy/morph/morphs/morphsqueeze.py
@@ -1,0 +1,36 @@
+from diffpy.morph.morphs.morph import LABEL_GR, LABEL_RA, Morph
+
+
+class MorphSqueeze(Morph):
+    """Squeeze the morph function.
+
+    This applies a polynomial to squeeze the morph non-linearly. The resulting
+    squeezed morph is interpolated to the (trimmed) target grid.
+    Only the overlapping region between the squeezed morph and the target
+    grid is used. The target is trimmed (or not) accordingly, and the final
+    outputs (morph and target) are returned on the same grid, defined by this
+    trimmed target range.
+
+    Configuration Variables
+    -----------------------
+    squeeze
+        list or array-like
+        Polynomial coefficients [a0, a1, ..., an] for the squeeze function.
+    """
+
+    # Define input output types
+    summary = "Squeeze morph by polynomial shift"
+    xinlabel = LABEL_RA
+    yinlabel = LABEL_GR
+    xoutlabel = LABEL_RA
+    youtlabel = LABEL_GR
+    parnames = ["squeeze"]
+
+    def morph(self, x_morph, y_morph, x_target, y_target):
+        Morph.morph(self, x_morph, y_morph, x_target, y_target)
+        if self.squeeze is None:
+            self.x_morph_out = self.x_morph_in
+            self.y_morph_out = self.y_morph_in
+            return self.xyallout
+
+        return self.xyallout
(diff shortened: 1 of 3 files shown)


src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -0,0 +1,28 @@
...
+
+class MorphSqueeze(Morph):
+    """Squeeze the morph function.
+
+    This applies a polynomial to squeeze the morph non-linearly. The morphed
+    data is returned on the same grid as the unmorphed data.
+
+    Configuration Variables
+    -----------------------
+    squeeze
```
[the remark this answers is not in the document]
Luiskitsu: Let's just have a list here, I think makes sense and is consistent.

tests/test_morphsqueeze.py
```diff
@@ -0,0 +1,60 @@
...
+    # Zeros and non-zeros, the full polynomial is applied
+    [0, 0.03, 0, -0.001],
+    # Testing zeros, expect no squeezing
+    [0, 0, 0, 0, 0, 0],
+]
+morph_target_grids = [
+    # UCs from issue 181: https://github.com/diffpy/diffpy.morph/issues/181
+    # UC2: Same grid
+    (np.linspace(0, 10, 101), np.linspace(0, 10, 101)),
+    # UC4: Target extends beyond morph
```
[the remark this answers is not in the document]
Luiskitsu: Yes, we have those cases covered. We have UC6 where the target is finer grid density and then UC8 where the morph is finer grid density. I'll update the comments in the test definitions to make this clearer, including specifying the grid density differences explicitly.

I think it's fine for the morph and target to have different grid densities in these tests, since we're not performing operations that compare the two arrays. That will become critical during refinement, where the grid densities will affect comparison between morph and target?.

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -0,0 +1,28 @@
+from diffpy.morph.morphs.morph import LABEL_GR, LABEL_RA, Morph
+
+
+class MorphSqueeze(Morph):
+    """Squeeze the morph function.
+
+    This applies a polynomial to squeeze the morph non-linearly. The morphed
+    data is returned on the same grid as the unmorphed data.
+
```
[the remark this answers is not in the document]
Luiskitsu: So I was following the architecture of other morphs. They have a docstring bellow the class which is the main description and then another docstring under the function which are few words. Should I keep it consistent? 
When I add an example in the docstring do I add it as code using >>>?

Luiskitsu:
I have edited the comments and docstrings as you mentioned. The test covers same grids as well as different grids for target vs morph inputs.  
When we are happy about the tests I will write the function code. 
For the tests, how do we choose the tolerance given to interpolation/extrapolation inaccuracies? For scattering UCs the squeezing parameters will be very small and therefore the tolerance can be very small. But for other user cases the squeezing parameters can be much larger. I guess we can also think it other way around, how do we choose the squeezing parameters for the tests? So far I am using values that are much larger than what we would expect in diffraction experiments.

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -9,9 +9,29 @@ class MorphSqueeze(Morph):
...
+        on.  The order of the polynomial is determined by the length of the
+        list.
+
+    Example
+    -------
+    >>> import numpy as np
+    >>> from numpy.polynomial import Polynomial
+    >>> from diffpy.morph.morphs.morphsqueeze import MorphSqueeze
+
+    >>> x_target = np.linspace(0, 10, 101)
```
sbillinge, line 24: Let's remove this block of text because it will confuse people.  This is us building a target, but the assumption is that the user already has a target, so none of this code is needed by people just to use the morph.

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -9,9 +9,29 @@ class MorphSqueeze(Morph):
...
+    >>> y_target = np.sin(x_target)
+    >>> x_morph = np.linspace(0, 10, 101)
+    >>> squeeze_coeff = [0.1, -0.01, 0.005]
+    >>> poly = Polynomial(squeeze_coeff)
+    >>> y_morph = np.sin(x_morph + poly(x_morph))
+
+    >>> morph = MorphSqueeze()
```
sbillinge, line 31: this block is what we are after.  We could add a few words about what the variables mean.  Maybe also add a couple of lines that show other attributes that the user can access from the `morph` instance.  For example things like
`x_morph_in = morph.x_morph_in` etc.  No need to be exhaustive, just give a few examples to help users.

btw, this looks great in general.  I am not sure if we have a "group standard" for these code blocks.  I think that it is possible to write htese actually so that they can be run by some automated testing codes but we don't really do that.  I just feel that, as a user, if there is a few lines of code showing how to use this, it can be super helpful.

tests/test_morphsqueeze.py
```diff
@@ -25,17 +25,17 @@
 ]
 morph_target_grids = [
     # UCs from issue 181: https://github.com/diffpy/diffpy.morph/issues/181
-    # UC2: Same grid
+    # UC2: Same range and same grid density
```
sbillinge, line 28: This is now fantastic!  I love these tests.  Future us will love you.

src/diffpy/morph/morphs/morphsqueeze.py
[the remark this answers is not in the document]
sbillinge: We don't have a group standard for this, but more recently I think we are leaning towards having the class-level docstring to be just a few words about the high level intent of the class and the docstring of the constructor to be have more detail.    I did look into this question at one point and formed an opinion, but I don't fully remember what I found.  It could have been when I was working on my beloved "DiffractionObject"s in diffpy.utils but I am not sure.  We could look there.     All documentation is better than lack of documentation, so the standards are just for us to figure out the most effective (user-useful) way to write the docs so that any time we spend on it has the most impact.

sbillinge (commented):
lovely!  One last comment on the docstring, but let's push on and implement the functionality!

Luiskitsu:
I have moved the main docstring below the function definition and the class-level docstring is the short description. I have also added comments on the example. 
When you push I will work on the functionality. For the tests, do we want to add a tolerance to account for the interpolation/extrapolation inaccuracy?

Luiskitsu:
Also, I just realized that some of my input 3rd order degree coefficients are too large and the x-array is not strictly increasing and loops back. When we work on the functionality, an idea is that we could add a warning statement when the x-array doesn't increase due to high values of high order polynomials and just return the input data and tell the user to change the initial values for the coefficients and run it again

Luiskitsu:
> > I think we are ready to write the code now. Regarding tolerances, the kind of behavior I would want is for the tolerances to be very tight on non-extrapolated regions and I could tolerate a slightly larger tolerance on extrapolated data (That is tagged as such as we discussed). It is probably a mistake to allow loose tolerances on all points to all tests to pass for a few points at the ends. For these unit tests, I think it just computes a squeezed function so I think it is not really extrapolating anything, just computing it over a wider range. maybe use 10-6 or sthg?

Sounds great! Should we push this first or should I keep working on the same PR? 
I can make the non-extrapolated regions have very tight tolerances. For these tests we are extrapolating for certain scenarios, when we unsqueeze squeezed data. When we squeeze unsqueezed data we don't extrapolate.

sbillinge:
btw, we should keep going on this PR.

Luiskitsu:
Our tests do involve extrapolation. The morph function uses `y_morph_actual = CubicSpline(x_squeezed, y_morph)(x_morph)`, which extrapolates when `x_morph` is outside `x_squeezed`. While our expected result is `y_morph_expected = np.sin(x_morph)`. Right now I am returning the indices before and after interpolation and comparing the interpolated region with a tight tolerance (1e-6). I can change the expected morph to match the spline exactly, but that would just be verifying the same code. 
So far 47 tests are passing and one test is failing, I think is due to larger interpolation error in one scenario. 
I've also adjusted the 3rd order coefficient (from 0.001 to 0.0001) to avoid any axis flipping which should not happen in any UC. Then I can add an assert for each extrapolated region but with larger tolerances than 1e-6. In some cases I expect the extrapolated error to be large (>1).

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -0,0 +1,73 @@
...
+        squeeze : list
+            The polynomial coefficients [a0, a1, ..., an] for the squeeze
+            function where the polynomial would be of the form
+            a0 + a1*x + a2*x^2 and so on.  The order of the polynomial is
+            determined by the length of the list.
+
+        Returns
+        -------
+            A tuple (x_morph_out, y_morph_out, x_target_out, y_target_out,
+            min_index, max_index) where the target values remain the same and
```
sbillinge, line 36: we don't need to return min and max index.  These can be just updated in place.  Also, these names are not so intuitive to me. maybe something like `extrap_index_low` and `extrap_index_high`?    Let's define these and set them to `None` when `MorphSqueeze` is instantiated.

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -0,0 +1,73 @@
...
+        Morph.morph(self, x_morph, y_morph, x_target, y_target)
+
+        squeeze_polynomial = Polynomial(self.squeeze)
+        x_squeezed = self.x_morph_in + squeeze_polynomial(self.x_morph_in)
+        self.y_morph_out = CubicSpline(x_squeezed, self.y_morph_in)(
+            self.x_morph_in
+        )
+        left_extrap = np.where(self.x_morph_in < x_squeezed[0])[0]
+        right_extrap = np.where(self.x_morph_in > x_squeezed[-1])[0]
+        min_index = left_extrap[-1] if left_extrap.size else None
```
sbillinge, line 71: `self.extrap_index_low` (or left) and don't return it.

Describe this variable in the constructor where it is first defined.

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -0,0 +1,73 @@
...
+        max_index = right_extrap[0] if right_extrap.size else None
+        return self.xyallout + (min_index, max_index)
```
sbillinge, line 73: apart from it being a bad idea in general, this breaks backwards compatibility.

tests/test_morphsqueeze.py
```diff
@@ -0,0 +1,78 @@
...
+    morph.squeeze = squeeze_coeffs
+    (
+        x_morph_actual,
+        y_morph_actual,
+        x_target_actual,
+        y_target_actual,
+        low_extrap_idx,
+        high_extrap_idx,
+    ) = morph(x_morph, y_morph, x_target, y_target)
+    if low_extrap_idx is None and high_extrap_idx is None:
```
sbillinge, line 62: this seems too complicated to me, and also does it test the extrapolation?  How about something like
```
if low_extrap_idx is None: 
    low_extrap_idx = 0
elif high_extrap_idx is None:
    high_extrap_idx = -1
assert np.allclose(y_morph_actual[low_extrap_index:high_extrap_idx], y_morph_expected[low_extrap_index:high_extrap_idx], atol=1e-6)
assert np.allclose(y_morph_actual[:low_extrap_index-1], y_morph_expected[:low_extrap_index-1], atol=1e-5)
assert np.allclose(y_morph_actual[high_extrap_index+1:], y_morph_expected[high_extrap_index+1:], atol=1e-5)
assert morph.low_extrap_idx == expected_low_extrap_idx
assert morph.high_extrap_idx == expected_high_extrap_idx
```

sbillinge (commented):
nearly there.  Few comments

Luiskitsu:
Hi Simon, thank you very much for the feedback! I have updated the test and the function accordingly. There are some tests that are failing and I am pretty sure is because there are cases where the extrapolation is large and inaccurate.

sbillinge:
> Hi Simon, thank you very much for the feedback! I have updated the test and the function accordingly. There are some tests that are failing and I am pretty sure is because there are cases where the extrapolation is large and inaccurate.

that's interesting in its own right.  What do you think is a good behavior?  One thing we can do in principle is to allow a user to specify a tolerance and alert them when an extrapolation exceeds that tolerance.  We could then have a default tolerance for the interpolation and for the extrapolation.    

That could actually be a nice "scientifically rigorous" solution.  However, it is making things more complicated.  Maybe a suggestion would be to 
1. decide tolerances we are comfortable with
2. write tests that pass with those tolerances (so reduce the size of the parameters)
3. write an issue for a future "feature" that implements the idea above and leave it for, well, the future.....

Luiskitsu:
I like your idea where the user can specify tolerance and we can warn them if it exceeds that tolerance for extrapolated values.
For the current tests I think having a tight tolerance of 1e-6 for interpolated values and a more loose tolerance of 1e-3 makes sense. I have made some of the high order coefficients smaller so that the tests pass. We could also write the tolerance as a percentage from the target data rather than just the subtraction.
I will add an issue for adding the input tolerance and tolerance warning

sbillinge:
super!!!!!  got it merged.  That was a huge job but we learned a lot about our morph functions so it was time well spent.

We should decide what is next.  We could 

1. generalize the scale and other morphs to have similar behavior
2. make a function morph that allows a user to specify a python function and the morph applies that to the y and/or x axes
3. implement the variable tolerance feature
4. something else

We started this so we could do the thing of morphing a low-Qmax F(Q) onto a high Qmax one, and I think that we want (2) to do that, then for the actual UC for the function to be the function in PDFgetX3 that applies the corrections to get the F(Q).....

Thoughts?

Luiskitsu:
Thanks again for all your mentoring!
I’m a bit confused regarding how we've been handling different Q-grids and Q-ranges in our tests for `MorphSqueeze`. I've been testing `diffpy.morph.morph_api` for stretching, and it looks like it can already handle different Q-grids and Q-ranges as inputs.
Looking at the tests in `test_morphstretch.py` there are no tests for cases with differing Q-ranges or Q-grids. Could it be that the handling of different x-axis has already been taken care of by `morphrgrid.py`?
We've been focusing on testing MorphSqueeze, but it seems like we may not have fully checked how the Q-range and Q-grid handling is implemented in the full morph chain function.

Luiskitsu:
Regarding the next steps, I think we should develop both a "multiplicative" morph and an "additive" morph to convert from I(q) to F(q) given an F(q) target (for our XFEL UC).
Lets say we have I(q) data from XFEL, and we want to morph it to match the F(q) target from synchrotron data.
The user should be able to specify the polynomial order for the additive component and a polynomial or another function (such as multiple gaussians) for the multiplicative component.
With this functionality, the morphing process will handle the entire conversion from I(q) to F(q) by fitting both the additive and multiplicative components. The result will be the morphed F(q) on the same grid as the input I(q).
For the additive component, I think using a polynomial makes the most sense as it is flexible. For the multiplicative component, we could give the user the option to specify whether they prefer gaussians or another function.
This morphing will allow us to get everything we need for our XFEL UC. We can make it more general for other UCs

sbillinge:
> Thanks again for all your mentoring! I’m a bit confused regarding how we've been handling different Q-grids and Q-ranges in our tests for `MorphSqueeze`. I've been testing `diffpy.morph.morph_api` for stretching, and it looks like it can already handle different Q-grids and Q-ranges as inputs. Looking at the tests in `test_morphstretch.py` there are no tests for cases with differing Q-ranges or Q-grids. Could it be that the handling of different x-axis has already been taken care of by `morphrgrid.py`? We've been focusing on testing MorphSqueeze, but it seems like we may not have fully checked how the Q-range and Q-grid handling is implemented in the full morph chain function. We might need to revisit our code so that it does not do the same thing twice? I think is fine if we keep it as it is, it won't affect negatively but I think the data is regridded before being refined.

I talked to Andrew and indeed, it is the case, according to him, that different grids and ranges are handled.  This of course makes things much easier.  

Regarding what to do, do we know how extrapolation and precision are handled in that case?

sbillinge:
> Regarding the next steps, I think we should develop both a "multiplicative" morph and an "additive" morph to convert from I(q) to F(q) given an F(q) target (for our XFEL UC).

I actually have a bit of an issue with this statement, though in the end we can/will implement this functionality.   What is my problem here? My problem is that this functionality is already implemented in PDFgetX3.   So why don't we just have PDFmorph call PDFgetX3 and refine the input parameters rpoly, qmaxinst, qmax and so on as the function variables?

To do this we just need a morph that applies a user-supplied python function that takes inputs and returns a morphed version of one of the inputs.   In principle, we need this on the y-axis and on the x-axis, though the y-axis one is the one we need initially.

Luiskitsu:
> > Thanks again for all your mentoring! I’m a bit confused regarding how we've been handling different Q-grids and Q-ranges in our tests for `MorphSqueeze`. I've been testing `diffpy.morph.morph_api` for stretching, and it looks like it can already handle different Q-grids and Q-ranges as inputs. Looking at the tests in `test_morphstretch.py` there are no tests for cases with differing Q-ranges or Q-grids. Could it be that the handling of different x-axis has already been taken care of by `morphrgrid.py`? We've been focusing on testing MorphSqueeze, but it seems like we may not have fully checked how the Q-range and Q-grid handling is implemented in the full morph chain function. We might need to revisit our code so that it does not do the same thing twice? I think is fine if we keep it as it is, it won't affect negatively but I think the data is regridded before being refined.
> 
> I talked to Andrew and indeed, it is the case, according to him, that different grids and ranges are handled. This of course makes things much easier.
> 
> Regarding what to do, do we know how extrapolation and precision are handled in that case?

I just checked `MorphStretch` and it is not really handling extrapolation:
Figure 18 (10)
We could include extrapolation in all the morphs correspondingly as well as handling extrapolation precision

Luiskitsu:
> > Regarding the next steps, I think we should develop both a "multiplicative" morph and an "additive" morph to convert from I(q) to F(q) given an F(q) target (for our XFEL UC).
> 
> I actually have a bit of an issue with this statement, though in the end we can/will implement this functionality. What is my problem here? My problem is that this functionality is already implemented in PDFgetX3. So why don't we just have PDFmorph call PDFgetX3 and refine the input parameters rpoly, qmaxinst, qmax and so on as the function variables?
> 
> To do this we just need a morph that applies a user-supplied python function that takes inputs and returns a morphed version of one of the inputs. In principle, we need this on the y-axis and on the x-axis, though the y-axis one is the one we need initially.

I didn't think about this as I thought that PDFgetX3 optimizes the parameters based on asymptotic behavior. If we can modify the cost function to refine against the F(q) target through the morphing process, that would align perfectly with our goals. I agree that using PDFgetX3 to refine parameters like rpoly and others to fit the target F(q) (e.g., synchrotron F(q)) makes a LOT more sense. Where can I learn more about the code for PDFgetX3?
I can start implementing this approach. Should we call this function MorphPDF?

sbillinge:
> I can start implementing this approach. Should we call this function MorphPDF?

we just need a general morph function as a first step.  Them to have this function be a method from PDFgetX3 we can either do that out of the box if that function is already available as part of the getX3 API, or we can see what mods are needed in getX3 to expose it.

Not sure what to call it.  Maybe `MorphFuncy` (and later `MorphFuncx`)?

The UC is
1. Luis wants to apply a user supplied python function to the y-coordinates of his morph data to make it agree with the target
2. Luis supplies the x-morph, y-morph and the function, where function is a python function that is a function of x
3. diffpy.morph applies the function to the morph and returns the morphed function

Luiskitsu:
This sounds like a fantastic plan! I am finishing experiments tomorrow at APS and will start working on this later on the week. I will create an issue for this.

PR #186, opened 2025-04-23
test/func: created a test and function for MorphFuncy

Luiskitsu:
I have created a draft for the `MorphFuncy` class and tested it using different functions with various parameters that I have defined using `pytest.parametrize`. I am currently defining the parameters for the functions as a dictionary, where each function has its specific parameters. I'm unsure if using a dictionary is the best decision. We can implement more options such as if the user does not provide a function to raise a warning or something similar.

### DIFF
--- a/src/diffpy/morph/morphs/morphfuncy.py
+++ b/src/diffpy/morph/morphs/morphfuncy.py
@@ -37,18 +37,22 @@ class MorphFuncy(Morph):
         -------
         Import the funcy morph function:
         >>> from diffpy.morph.morphs.morphfuncy import MorphFuncy
+
         Define or import the user-supplied transformation function:
         >>> def sine_function(x, y, amplitude, frequency):
         >>>     return amplitude * np.sin(frequency * x) * y
+
         Provide initial guess for parameters:
         >>> parameters = {'amplitude': 2, 'frequency': 2}
+
         Run the funcy morph given input morph array (x_morph, y_morph)
         and target array (x_target, y_target):
         >>> morph = MorphFuncy()
         >>> morph.function = sine_function
         >>> morph.parameters = parameters
-        >>> x_morph_out, y_morph_out, x_target_out, y_target_out = morph(
+        >>> x_morph_out, y_morph_out, x_target_out, y_target_out = morph.morph(
         ...     x_morph, y_morph, x_target, y_target)
+
         To access parameters from the morph instance:
         >>> x_morph_in = morph.x_morph_in
         >>> y_morph_in = morph.y_morph_in
@@ -61,5 +65,4 @@ class MorphFuncy(Morph):
         self.y_morph_out = self.function(
             self.x_morph_in, self.y_morph_in, **self.parameters
         )
-
         return self.xyallout
--- a/src/diffpy/morph/morphs/morphfuncy.py
+++ b/src/diffpy/morph/morphs/morphfuncy.py
@@ -1,0 +1,65 @@
+from diffpy.morph.morphs.morph import LABEL_GR, LABEL_RA, Morph
+
+
+class MorphFuncy(Morph):
+    """Apply the user-supplied Python function to the y-coordinates of the
+    morph data"""
+
+    # Define input output types
+    summary = "Apply a Python function to the y-axis data"
+    xinlabel = LABEL_RA
+    yinlabel = LABEL_GR
+    xoutlabel = LABEL_RA
+    youtlabel = LABEL_GR
+
+    def morph(self, x_morph, y_morph, x_target, y_target):
+        """General morph function that applies a user-supplied function to the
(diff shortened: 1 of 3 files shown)


tests/test_morphfuncy.py
```diff
@@ -0,0 +1,73 @@
+import numpy as np
+import pytest
+
+from diffpy.morph.morphs.morphfuncy import MorphFuncy
+
```
sbillinge, line 5: I think these functions would be better with single blank lines between them.

tests/test_morphfuncy.py
```diff
@@ -0,0 +1,73 @@
...
+    x_morph = np.linspace(0, 10, 101)
+    y_morph = np.sin(x_morph)
+    x_target = x_morph.copy()
+    y_target = y_morph.copy()
+    x_morph_expected = x_morph
+    y_morph_expected = expected_function(x_morph, y_morph)
+    morph = MorphFuncy()
+    morph.function = function
+    morph.parameters = parameters
+    x_morph_actual, y_morph_actual, x_target_actual, y_target_actual = (
```
sbillinge, line 67: Is there a reason you have the function call inside parens?  It should work the same way without these and this is somewhat unconventional and therefore distracting?

src/diffpy/morph/morphs/morphfuncy.py
```diff
@@ -0,0 +1,65 @@
...
+            A tuple (x_morph_out, y_morph_out, x_target_out, y_target_out)
+            where the target values remain the same and the morph data is
+            transformed according to the user-specified function and parameters
+            The morphed data is returned on the same grid as the unmorphed data
+
+        Example
+        -------
+        Import the funcy morph function:
+        >>> from diffpy.morph.morphs.morphfuncy import MorphFuncy
+        Define or import the user-supplied transformation function:
```
sbillinge, line 40: I wonder if this would be easier to read with judiciously placed blank lines, maybe after each code block, but play around till it looks nice?  I am not sure if there is a docstring standard for this, in which case follow it, but if not, then I think a few blank lines could increase readability.

sbillinge (commented):
wow, fantastic.  I few tiny  cosmetic comments, but you nailed it in one shot as far as I can tell!

tests/test_morphfuncy.py
Luiskitsu: I am trying to delete these but every time I commit the files are modified by the hooks. I think this is the pre-commit?

tests/test_morphfuncy.py
Luiskitsu: This is same as comment above, this is modified by the hooks.

Luiskitsu:
Thanks Simon! This is your mentoring ... after MorphSqueeze I did learn a lot. 
As I mentioned in the comments, a lot of the cosmetic issues are modified by the hooks. Not sure how I can change these.

tests/test_morphfuncy.py
sbillinge: OK,, we can let it slide.  It is because normally we want the functions well separated, but wehn we are defining a bunch of them for testing we just want them kind of closed up, but it is not a big deal.

Just fix the docstring and I can merge this.

sbillinge (commented):
pls see inline

Luiskitsu:
@sbillinge  should I work on MorphFuncx? Do we want to do the same as MorphFuncy but applied to the x-axis and then return the morphed data in the same grid as the input morph data? It would also be good if I could see the code for PDFgetX3, I can't find it in github.

sbillinge:
> @sbillinge should I work on MorphFuncx? Do we want to do the same as MorphFuncy but applied to the x-axis and then return the morphed data in the same grid as the input morph data? It would also be good if I could see the code for PDFgetX3, I can't find it in github.

I do want this functionality in the fullness of time, and it should be relatively straightforward because we have the squeeze morph that uses a "particular" function which was the polynomial.  So it should be relatively quick, almost a copy-paste.   

However, I would rather push next to see if we can figure out how to use a getX3 function to do the moprh onto the synchrtron data......

Luiskitsu:
> > @sbillinge should I work on MorphFuncx? Do we want to do the same as MorphFuncy but applied to the x-axis and then return the morphed data in the same grid as the input morph data? It would also be good if I could see the code for PDFgetX3, I can't find it in github.
> 
> I do want this functionality in the fullness of time, and it should be relatively straightforward because we have the squeeze morph that uses a "particular" function which was the polynomial. So it should be relatively quick, almost a copy-paste.
> 
> However, I would rather push next to see if we can figure out how to use a getX3 function to do the moprh onto the synchrtron data......

Sounds great to me!
Where can I find the documentation for getX3 functions?

sbillinge:
let me go and dig that up.  This code is not open source.....

PR #190, opened 2025-04-29
Enable refinement of lists and dictionary input parameters

Luiskitsu:
I have modified the `refine.py` to allow refinement of scalars, lists, and dictionaries (before it was only scalars). I have done this by flattening lists and dictionaries accordingly and then reconstructing and regrouping the lists or dictionaries after refinement.
I have also added `MorphSqueeze` and `MorphFuncy` into `morph_api.py` and `_init_.py`. I have finally tested that we can now refine `MorphSqueeze` and `MorphFuncy` and I added these tests in `test_morph_func.py` where there are tests for other morphs as well. The tests are successful and therefore now we can refine squeeze and funcy. I am attaching some figures from the tests here. Finally, I also run all the tests for all the diffpy.morph, such as `test_refine.py`, etc., to make sure everything is working properly and they all passed. 
Refinement test for scale + squeeze morph with same x-grid:
image
Refinement test for scale + squeeze morph with different x-grid:
image
Refinement test for `MorphFuncy` using a linear function that is applied to a sine wave:
image
Note that I also compare the refined parameters to the expected ones and is working well.

### DIFF
--- a/src/diffpy/morph/morph_api.py
+++ b/src/diffpy/morph/morph_api.py
@@ -40,7 +40,7 @@ _morph_step_dict = dict(
     ],
     qdamp=morphs.MorphResolutionDamping,
     squeeze=morphs.MorphSqueeze,
-    parameters=morphs.MorphFuncy,
+    funcy=morphs.MorphFuncy,
 )
 _default_config = dict(
     scale=None,
@@ -49,7 +49,7 @@ _default_config = dict(
     baselineslope=None,
     qdamp=None,
     squeeze=None,
-    parameters=None,
+    funcy=None,
 )
 
 
@@ -144,7 +144,7 @@ def morph(
             - 'baselineslope'
             - 'qdamp'
             - 'squeeze'
-            - 'parameters'
+            - 'funcy'
 
     Returns
     -------
@@ -207,7 +207,7 @@ def morph(
         if k == "smear":
             [chain.append(el()) for el in morph_cls]
             refpars.append("baselineslope")
-        elif k == "parameters":
+        elif k == "funcy":
             morph_inst = morph_cls()
             morph_inst.function = rv_cfg.get("function", None)
             if morph_inst.function is None:
--- a/src/diffpy/morph/morph_api.py
+++ b/src/diffpy/morph/morph_api.py
@@ -143,6 +143,8 @@ def morph(
             - 'smear'
             - 'baselineslope'
             - 'qdamp'
+            - 'squeeze'
+            - 'parameters'
 
     Returns
     -------
--- a/src/diffpy/morph/morph_api.py
+++ b/src/diffpy/morph/morph_api.py
@@ -39,9 +39,17 @@ _morph_step_dict = dict(
         morph_helpers.TransformXtalRDFtoPDF,
     ],
     qdamp=morphs.MorphResolutionDamping,
+    squeeze=morphs.MorphSqueeze,
+    parameters=morphs.MorphFuncy,
 )
 _default_config = dict(
-    scale=None, stretch=None, smear=None, baselineslope=None, qdamp=None
+    scale=None,
+    stretch=None,
+    smear=None,
+    baselineslope=None,
+    qdamp=None,
+    squeeze=None,
+    parameters=None,
 )
 
 
@@ -197,6 +205,14 @@ def morph(
         if k == "smear":
             [chain.append(el()) for el in morph_cls]
             refpars.append("baselineslope")
+        elif k == "parameters":
+            morph_inst = morph_cls()
+            morph_inst.function = rv_cfg.get("function", None)
+            if morph_inst.function is None:
+                raise ValueError(
+                    "Must provide a 'function' when using 'parameters'"
+                )
+            chain.append(morph_inst)
         else:
             chain.append(morph_cls())
         refpars.append(k)
--- a/src/diffpy/morph/refine.py
+++ b/src/diffpy/morph/refine.py
@@ -26,8 +26,6 @@ from scipy.stats import pearsonr
 class Refiner(object):
     """Class for refining a Morph or MorphChain.
 
-    This is provided to allow for custom residuals and refinement algorithms.
-
     Attributes
     ----------
     chain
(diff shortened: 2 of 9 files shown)


sbillinge:
I commented on the issues.  Since you got this working I am inclinde to review it and merge it, but I would still like a discussion if this is the solution we want.  If it is a workaround that needs a better fix we can create an issue and attach it to a future release maybe.

src/diffpy/morph/morph_api.py
```diff
@@ -39,9 +39,17 @@
...
 )
 _default_config = dict(
-    scale=None, stretch=None, smear=None, baselineslope=None, qdamp=None
+    scale=None,
+    stretch=None,
+    smear=None,
+    baselineslope=None,
+    qdamp=None,
+    squeeze=None,
+    parameters=None,
```
sbillinge, line 52: is there a reason we are not calling this `funcy`? or `funcy_parameters`?  That seems to follow the previous pattern better.

src/diffpy/morph/refine.py
```diff
@@ -51,12 +51,27 @@ def __init__(self, chain, x_morph, y_morph, x_target, y_target):
...
         self.residual = self._residual
+        self.flat_to_grouped = {}
         return
 
     def _update_chain(self, pvals):
         """Update the parameters in the chain."""
-        pairs = zip(self.pars, pvals)
-        self.chain.config.update(pairs)
+        updated = {}
+        for idx, value in enumerate(pvals):
```
sbillinge, line 60: This is a bit hard to follow.  I think the code is fine, I just wonder if there is a more elegant and readable way of doing this.

I guess the basic structure "before" was that we had two ordered lists of equal length containing parameter names(?) and values.  Apart from this structure being a bit brittle and possibly not optimal, but ok, np.  I guess the task here then is that we introduce parameters that are not singles but can be, themselves lists or dicts.  If they are sometimes lists and sometimes dicts are shooting ourselves in the foot because we actually control this ourselves, so we could maybe decide on what the best structure is and then stick to that.  If we pick a dict, which seems to be the most logical, then there is no different between singles, doubles etc..  So I suggest that we think about this, but maybe change the basic type of our "parameter" objects.  I would vote for dicts. 

to be clear, I am ok parsing them out into zipped lists for passing to other parts of the code if that is what they are expecting, but just storing them in the Morph object as a dict seems to me to make the most sense.

src/diffpy/morph/refine.py
```diff
@@ -118,11 +133,29 @@ def refine(self, *args, **kw):
         if not self.pars:
             return 0.0
 
-        initial = [config[p] for p in self.pars]
+        # Build flat list of initial parameters and flat_to_grouped mapping
+        initial = []
+        self.flat_to_grouped = {}
+
+        for p in self.pars:
```
sbillinge, line 140: my suggested refactor probably gets rid of this too....so there is a lot of technical debt we remove with this refactor if we decide to do it.

tests/test_morph_func.py
```diff
@@ -101,3 +101,47 @@ def test_smear_with_morph_func():
     assert np.allclose(y0, y1, atol=1e-3)  # numerical error -> 1e-4
     # verify morphed param
     assert np.allclose(smear, morphed_cfg["smear"], atol=1e-1)
+
+
+def test_squeeze_with_morph_func():
```
sbillinge, line 106: These tests look awesome!

sbillinge (commented):
please see my comments.  It looks really great as a general matter.

src/diffpy/morph/morph_api.py
Luiskitsu: I will change this to `funcy` to follow the pattern!

src/diffpy/morph/refine.py
Luiskitsu: But I agree with you, from the user perspective it is better to be consistent and will be better to having all dicts

Luiskitsu: If we want to have squeeze as a list I can do the corresponding modifications to the code and the tests. The `refine.py` will be simplified to 
```
def _update_chain(self, pvals):
    """Update the parameters in the chain, supporting only scalars and dictionaries."""
    updated = {}
    for idx, value in enumerate(pvals):
        param, subkey = self.flat_to_grouped[idx]
        if subkey is None:
            # Scalar parameter
            updated[param] = value
        else:
            # Dictionary
            if param not in updated:
                updated[param] = {}
            updated[param][subkey] = value

    self.chain.config.update(updated)
    return
```

src/diffpy/morph/refine.py
Luiskitsu: If we just had scalars and dicts this would be simiplified to:
```
for p in self.pars:
    val = config[p]
    if isinstance(val, dict):
        for k, v in val.items():
            initial.append(v)
            self.flat_to_grouped[len(initial) - 1] = (p, k)
    else:
        initial.append(val)
        self.flat_to_grouped[len(initial) - 1] = (p, None)
```

src/diffpy/morph/refine.py
sbillinge: internals of the code and user experience are two different things.  We can wrap anything.  The basic conversation is "what is the best type for the basic parameter data object?"  We would often define a container class to carry these things around, something like `class MorphParameters` but since it is basically a set we could just use a  dictionary.  There would be no "scalar" parameter.  A scalar parameter would be a dictionary with one key.

src/diffpy/morph/refine.py
sbillinge: no scalars, just dicts.

src/diffpy/morph/refine.py
Luiskitsu: Sounds good! So far, I’ve modified `MorphSqueeze` so that its parameter `squeeze` is now represented as a dictionary—for example: `squeeze = {"a0":0, "a1":0.01, "a2":0.0001, "a3":0.001}`
If we want to generalize this approach, I could go ahead and update all the morphs, tests and other functions, so that all parameters that are scalars are stored as dictionaries—for instance: `scale = {"scale": 0.1}` 
With this change the refine code will be reduced to: 
```
def _update_chain(self, pvals):
    """Update the parameters in the chain, assuming all parameters are dictionaries."""
    updated = {}
    for idx, value in enumerate(pvals):
        param, subkey = self.flat_to_grouped[idx]
        updated.setdefault(param, {})[subkey] = value
    self.chain.config.update(updated)
    return
```
However, I’m also wondering if it's worth preserving scalars as-is for simplicity and only using dictionaries where multiple sub-parameters are required. That would avoid the somewhat inelegant repetition of having to write `scale = {"scale": 0.1}`.
Happy to implement either approach. We can also discuss this tomorrow and choose what direction we want to take.

Luiskitsu: What if we keep scalars and have scalars and dictionaries and create an issue for a near future? A lot of the code that was already written for the other morphs and chains will need to be modified, so I am not sure how quick of a fix is this

PR #199, opened 2025-06-10
skpkg: setup CI after migrating tests, src, requirements, and .github folder

zmx27:


zmx27:
@sbillinge pre-commit is passing locally for me, but I can't seem to make it pass CI. To me, it seems to be because I haven't used `git add` on the files where the errors are occurring (in my local repo, they would have been deleted, etc), but I'm not completely sure. Could you help me debug?

pyproject.toml
```diff
@@ -32,9 +32,6 @@ classifiers = [
         'Topic :: Scientific/Engineering :: Chemistry',
 ]
 
-[project.scripts]
```
sbillinge, line 35: this block we probably need to keep.  this is the diffpy.morph entry point.

requirements/conda.txt
```diff
@@ -1,5 +1 @@
 numpy
-scipy
```
sbillinge, line 2: I am pretty sure we don't want to delete these.  These are the requirements.

requirements/pip.txt
```diff
@@ -1,5 +1 @@
 numpy
-scipy
```
sbillinge, line 2: same here. Keep these

src/diffpy/__init__.py
```diff
@@ -1,25 +1,14 @@
...
 # File coded by: Billinge Group members and community contributors.
 #
 # See GitHub contributions for a more detailed list of contributors.
-# https://github.com/diffpy/diffpy.pdfmorph/graphs/contributors
+# https://github.com/diffpy/diffpy.morph/graphs/contributors
 #
 # See LICENSE.rst for license information.
 #
 ##############################################################################
-
```
sbillinge, line 15: I expect we need to keep this whole code-block too.

src/diffpy/morph/__init__.py
```diff
@@ -1,28 +1,23 @@
...
 # https://github.com/diffpy/diffpy.morph/graphs/contributors
 #
 # See LICENSE.rst for license information.
 #
 ##############################################################################
-
-"""Tools for manipulating and comparing PDF profiles"""
-
-# key used when saving multiple morphs
-__save_morph_as__ = "save_morph_as"
```
sbillinge, line 19: This code block needed too but it would go below the docstring.

sbillinge (commented):
please see comments.  Whenever you see things in red (deleted) look very carefully as they will be lost when this PR is merged.  It is quite hard to get them back (not that hard but a hassle).  If you are not sure, put a comment/question, but I went through and left my comments, so for now, just respond to those. I did it fairly carefully.

PR #201, opened 2025-06-10
skpkg: setup CI after migrating tests, src, requirements, and .github folder

zmx27:

PR #202, opened 2025-06-11
skpkg: add configuration files for pre-commit, readthedocs, codecov

zmx27:
#201 needs to be closed first


src/diffpy/__init__.py
```diff
@@ -12,3 +12,13 @@
 # See LICENSE.rst for license information.
 #
 ##############################################################################
+
```
sbillinge: per convo with @bobleesj this block can be deleted so (apart from the comments with the copyright statement) this file should be empty.

sbillinge (commented):
ok, this looks ok.  I am not sure if we want the blocks with setup.cfg and the socioeconomic etc. that you added back in the last commit.  Why did you add htat back?  Was it not passing pre-commit without those?  They are small things though, it doesn't really matter that much if they are there but not needed (just a bit of technical debt to carry forward to the future).  But let me know why you added them back (as well as addressing the one inline issue).

This is marked as draft.  Is there a reason for this?  It is not finished on your side?

zmx27:
Sorry, I forgot to mark it as review yesterday. Regarding setup.cfg, it seemed quite important for normal functionality, but I deleted it if it's not required. As for the socioeconomic block, I thought that it was needed to pass pre-commit, but it ended up not being required so I deleted it too. @sbillinge ready for review after addressing your concerns and comments

PR #219, opened 2025-06-23
Add test for prolate spheroid

### LINKED ISSUE #86: add test for prolate spheroid

Sparks29032:
Added test data for prolate spheroid similar to that of oblate spheroid from [this commit](https://github.com/diffpy/diffpy.morph/commit/13a2006c28199a2ed2b434eabd1946e8bbc19bc7). Same function is used for all spheroid morphs. Closes #86.

Not user-facing, so no news item added.


Sparks29032:
> > Not user-facing, so no news item added.
> 
> no problem. The preferred workflow for this is to install scikit-package in your env and run `package add no-news -m "why not needed"` (though double check the syntax.

Got it.

PR #220, opened 2025-06-23
Rename PDF-specific smear option to `--smear-pdf`

### LINKED ISSUE #210: Remove PDF-dependence on smear morph
### Problem

Smear morph right now only applies to PDF as it first converts the function to RDF before applying the smear. Change its functionality to have general applicability.

### Proposed solution
Rename current `smear` morph to `smear-pdf`. Remove the conversion between the functions.

Sparks29032:
Closes #210.

The `--smear-pdf` option first converts the PDF to an RDF before applying the smear. The `--smear` option does not do such transformation. If both are enabled, the `--smear-pdf` option takes precedence.


Sparks29032:
> @Sparks29032 I am afraid this one has a conflict.

Ah, probably from me fixing a typo on another PR. Should be resolved now.

tests/test_morphapp.py
```diff
@@ -255,6 +256,68 @@ def test_morphsequence(self, setup_morphsequence):
...
+        # Variances add, and 3^2+4^2=5^2
+        assert pytest.approx(abs(smear_results["smear"])) == 4.0
+        assert pytest.approx(smear_results["Rw"]) == 0.0
+
+        # PDF-specific smear (should activate baseline slope)
+        (opts, _) = self.parser.parse_args(
+            [
+                "--smear",
+                "100",
+                "--smear-pdf",
```
Sparks29032, line 309: This code block also tests the default behavior if a user inputs both morphs. The pdf smear takes precedence.

Sparks29032:
I thought these tests were already migrated from unittest to Pytest in #83. These tests seem to follow the Pytest [documentation](https://docs.pytest.org/en/stable/getting-started.html).

sbillinge:
> Thanks for merging! I thought these tests were already migrated from unittest to Pytest in #83. These tests seem to follow the Pytest [documentation](https://docs.pytest.org/en/stable/getting-started.html).

maybe I didn't check well enough.  It looked as if it was indented and part of a class and not a standalone function.

Sparks29032:
They are classes, but I thought this is supported and encouraged by https://docs.pytest.org/en/stable/getting-started.html.

sbillinge:
> They are classes, but I thought this is supported and encouraged by https://docs.pytest.org/en/stable/getting-started.html. I think I need the python tests presentation lol.

Ah, I see. I don't really see any benefit unless tests are sharing a bunch of code or marks, but mostly I just assumed they were unittest tests (which are typically the only tests we write in classes.....)

Issue #207, opened 2025-06-16
Documentation for morphsqueeze morphyfunc


Sparks29032:
### Problem

Need documentation for the functions. Add to `quickstart.rst` and `tutorialfiles.rst`. Any data used (if any) can be added to the `tutorial` directory in root.

Sparks29032:
Closed by #226

PR #226, opened 2025-06-30
Added MorphSqueeze and MorphFuncy tutorials

Luiskitsu:
Hi Simon, I have added the tutorials and also we found a indentation issue on docstring. For the MorphSqueeze we also want to be able to get the extrapolation values index and so far is only supported in python interface. Andrew will incorporate this later in the CLI.


Luiskitsu:
closes #207

PR #227, opened 2025-06-30
Edge cases and error handling for the squeeze morph

Sparks29032:
Throw error when invalid input is given (e.g. non-numeric list).

Indicate to use that duplicated/repeated commas and trailing commas will be removed: e.g. `0,,1,` becomes `0,1`.

### DIFF
--- a/tests/test_morphapp.py
+++ b/tests/test_morphapp.py
@@ -198,6 +198,18 @@ class TestApp:
         with pytest.raises(SystemExit):
             multiple_targets(self.parser, opts, pargs, stdout_flag=False)
 
+        # Pass a non-float list to squeeze
+        (opts, pargs) = self.parser.parse_args(
+            [
+                f"{nickel_PDF}",
+                f"{nickel_PDF}",
+                "--squeeze",
+                "1,a,0",
+            ]
+        )
+        with pytest.raises(SystemExit):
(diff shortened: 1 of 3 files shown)


tests/test_morphapp.py
```diff
@@ -198,6 +198,18 @@ def test_parser_systemexits(self, setup_parser):
...
+        # Pass a non-float list to squeeze
+        (opts, pargs) = self.parser.parse_args(
+            [
+                f"{nickel_PDF}",
+                f"{nickel_PDF}",
+                "--squeeze",
+                "1,a,0",
+            ]
+        )
+        with pytest.raises(SystemExit):
```
sbillinge, line 210: why are we using sys.exit and not a proper exception?  In general, this is to be avoided as it bypasses all the nice exception handling you get for free in Python?

sbillinge (commented):
one small question

tests/test_morphapp.py
Sparks29032: The line `parser.error(f"{coeff} could not be converted to float.")` is calling the in-built option parser function to handle the error once we detect a ValueError on the Python side.

Sparks29032: Hmm, but we should ensure that the test is capturing this particular error, and not necessarily system exiting elsewhere. Let me put a new commit in a bit.

PR #231, opened 2025-07-09
Update numerical file comparisons in tests

Sparks29032:
Create a shared function for numerical file comparisons.

### DIFF
--- a/tests/test_morphio.py
+++ b/tests/test_morphio.py
@@ -50,6 +50,20 @@ def isfloat(s):
     return False
 
 
+def compare_numeric(file1, file2):
+    """Assert that two files have (approximately) the same numerical
+    values."""
+    for f1_row, f2_row in zip(file1, file2):
+        f1_arr = f1_row.split()
+        f2_arr = f2_row.split()
+        assert len(f1_arr) == len(f2_arr)
+        for idx, _ in enumerate(f1_arr):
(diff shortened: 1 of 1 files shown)


tests/test_morphio.py
```diff
@@ -50,6 +50,20 @@ def isfloat(s):
     return False
 
 
+def compare_numeric(file1, file2):
```
sbillinge, line 53: this should probably have a name like `are_same(file1, file2)` since it returns True if they are the same and False if they are not.  Something like that.

tests/test_morphio.py
```diff
@@ -149,7 +163,7 @@ def test_morph_outputs(self, setup, tmp_path):
                 with open(test_saving_verbose.joinpath(file)) as tf:
                     generated = filter(ignore_path, gf)
                     target = filter(ignore_path, tf)
-                    assert all(x == y for x, y in zip(generated, target))
+                    compare_numeric(generated, target)
```
sbillinge, line 166: can we call the options `expected` and `actual` to make the test more intelligible

sbillinge (commented):
looks good.  Couple of small comments

tests/test_morphio.py
Sparks29032: Resolved

Issue #236, opened 2025-07-11
Replace all references of PDF to general 1D function language


Sparks29032:
This includes:

- [ ] Plot labels
- [ ] Function docstrings
- [ ] Save file headers
- [ ] etc.

PR #237, opened 2025-07-11
Add option to save difference curve

Sparks29032:
Very useful for experiments when you are interested in the difference signal. `diffpy.morph` has a very robust way of taking the difference on an intersecting grid as well

### DIFF
--- a/src/diffpy/morph/morphpy.py
+++ b/src/diffpy/morph/morphpy.py
@@ -45,6 +45,7 @@ def __get_morph_opts__(parser, scale, stretch, smear, plot, **kwargs):
         "apply",
         "reverse",
         "diff",
+        "get-diff",
     ]
     opts_to_ignore = ["multiple-morphs", "multiple-targets"]
     for opt in opts_storing_values:
--- a/tests/test_morphpy.py
+++ b/tests/test_morphpy.py
@@ -75,7 +75,7 @@ class TestMorphpy:
             "addpearson": False,
             "apply": False,
             "reverse": False,
-            "diff": False,
+            "get_diff": False,
             "multiple_morphs": False,
             "multiple_targets": False,
         }


docs/source/morphpy.rst
```diff
@@ -74,6 +74,10 @@ General Parameters
 
 save: str or path
     Save the morphed function to a the file passed to save. Use '-' for stdout.
+get_diff: bool
```
sbillinge, line 77: I wonder if we want to do either/or here.   A clearer logic would be that diff is saved if this is true, not saved if it is false, and also the case for the morph function.  So TT would save both.  To keep the API cleaner, we could have defaults of TF for morph and diff (the current behavior).

src/diffpy/morph/morphapp.py
```diff
@@ -81,14 +81,27 @@ def custom_error(self, msg):
...
             "--save-names-file."
         ),
     )
+    parser.add_option(
+        "--diff",
+        "--get-diff",
+        dest="get_diff",
+        action="store_true",
+        help=(
+            "Save the difference curve rather than the manipulated function.\n"
```
sbillinge, line 99: if we do update the logic, update this help here too.

src/diffpy/morph/morphapp.py
```diff
@@ -99,12 +112,12 @@ def custom_error(self, msg):
     parser.add_option(
         "--rmin",
         type="float",
-        help="Minimum r-value to use for PDF comparisons.",
+        help="Minimum r-value (abscissa) to use for function comparisons.",
```
sbillinge, line 115: "r" might be a bit confusing in our more general world of functions.  Not sure what to change it to (and let's just make an issue, not do it on this PR0.   "x" is  widely used by scientists.  "minimum value on the absisca" might be mathematically more correct?  Anyway, just a note to make an issue.....

src/diffpy/morph/morphapp.py
```diff
@@ -419,9 +432,9 @@ def custom_error(self, msg):
             "using a serial file NAMESFILE. The format of NAMESFILE should be "
             "as follows: each target PDF is an entry in NAMESFILE. For each "
             "entry, there should be a key {__save_morph_as__} whose value "
-            "specifies the name to save the manipulated PDF as. An example "
```
sbillinge, line 422: another issue to make as claning before release (not now) would be to do a global search for PDF and root it out wherever it survives.   Like a weed......

src/diffpy/morph/morphapp.py
Sparks29032: Yep, I noted this in #236

PR #244, opened 2025-07-15
Change kwargs input from dict to any

Sparks29032:
This is a docstring update. Prior to this, we get a warning if any of our kwarg entries are not dicts. The kwargs entry is supposed to denote the possible entry types, not that kwargs is a dict.


Sparks29032:
Merging this since it's a docstring update.

Issue #241, opened 2025-07-13
Raise exceptions for known errors


Sparks29032:
- [ ] On `squeeze`, if the polynomial squeeze causes the function grid to be non-strictly-increasing, an error is thrown. Let's wrap the `CubicSpline` call in `morphsqueeze.py` with a `try` `catch` to throw a custom warning to the user.
- [ ] If the grids of the morphed and target functions after the first iterations are too far separated (e.g. there is no overlap, or only one point overlap), then it is possible for the number of parameters to be smaller than the number of points we are regressing to fit over. This causes an error. In `refine.py` in the `_residual` function, let's throw a custom error to the user if `len(rvec) < len(pvals)`.

ycexiao:
The situation described in the first unchecked item will be addressed differently in #256.

sbillinge:
Thanks @ycexiao . Per your comment I am closing this as completed or moved to #256

PR #248, opened 2025-08-18
raise `ValueError` if `x_squeezed` is not strictly increasing

ycexiao:

Address the first item in #241.

raise ValueError if `x_squeezed` is not strictly increasing.

Please check the implementation.

`if` instead of `try` and `except` is used since it is more specific to the failed case. Please see which one do we want to use.

### DIFF
--- a/src/diffpy/morph/morphs/morphsqueeze.py
+++ b/src/diffpy/morph/morphs/morphsqueeze.py
@@ -87,7 +87,8 @@ class MorphSqueeze(Morph):
         if not strictly_increasing_x:
             raise ValueError(
                 "Computed squeezed x is not strictly increasing. "
-                "Please change the input x_morph or the squeeze "
+                "The squeezed morph is only intended for small polynomial "
+                "stretches. Please decrease the magnitude of the polynomial "
                 "coefficients."
             )
 


src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -78,6 +78,13 @@ def morph(self, x_morph, y_morph, x_target, y_target):
         coeffs = [self.squeeze[f"a{i}"] for i in range(len(self.squeeze))]
         squeeze_polynomial = Polynomial(coeffs)
         x_squeezed = self.x_morph_in + squeeze_polynomial(self.x_morph_in)
+        strictly_increasing_x = (np.diff(x_squeezed) > 0).all()
+        if not strictly_increasing_x:
+            raise ValueError(
+                "Computed squeezed x is not strictly increasing. "
+                "Please change the input x_morph or the squeeze "
```
sbillinge, line 85: If possible let's have a more helpfule message here.  I am a user.  OK, I need to change those things but what do I change them to?  How do I change them?

ycexiao:
> Is there an alternative to CubicSpline that lets us interpolate onto a non-increasing grid? 

I think the motive to use a strictly increasing grid is to make sure there is only one function value per `r` input. Non-monotonicity of a continuous function itself implies there are duplicated function values, which is `x_squeezed` in here. And duplicated `x_squeezed` should be handled as an Error.

Sparks29032:
Non-monotonicity does not necessarily imply duplicated values on a discrete grid.

My main concern is that this morph often gives this error during refinement on a dense grid since switching two adjacent function values around does not greatly change the Rw, hence giving non-monotonic squeeze values.

Is there a way that we can enforce monotonicity during refinement? Or can we correct for non-monotonicity somehow?

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -83,6 +83,15 @@ def morph(self, x_morph, y_morph, x_target, y_target):
         coeffs = [self.squeeze[f"a{i}"] for i in range(len(self.squeeze))]
         squeeze_polynomial = Polynomial(coeffs)
         x_squeezed = self.x_morph_in + squeeze_polynomial(self.x_morph_in)
+        strictly_increasing_x = (np.diff(x_squeezed) > 0).all()
+        if not strictly_increasing_x:
+            raise ValueError(
+                "Computed squeezed x is not strictly increasing. "
+                "The squeezed morph is only intended for small polynomial "
+                "stretches. Please decrease the magnitude of the polynomial "
```
sbillinge, line 91: this still seems pretty cryptic to me.  If I got this message as a user I wouldn't know what to change if I ran again.  there is agreat discussion on the comment thread which is way clearer.  Why don't we capture some of that here.  It is ok if the message is long.  @Sparks29032 please could you draft something.  Tell the whole story, including the "trick" of getting good starting values with `squeeze` and `hshift` and using for `a0` and `a1`.

Btw, the fact that this works seems to suggest that it is often a convergence issue and not, in those cases, a fundamental issue, i.e., the regression is stuck in a local minimum far from the right solution when this error happens, so we can mention that too?

Sparks29032: @sbillinge The story I would tell is only empirical from the limited times I've used this morph:
```
Error: The polynomial applied by the squeeze morph has resulted in a grid that is no longer strictly increasing. A strictly increasing grid is required for diffpy.morph to compute the morphed function through cubic spline interpolation. Here are some suggested methods to resolve this:
(1) Please decrease the order of your polynomial and try again.
(2) If you are using initial guesses of all 0, please ensure your objective function only requires a small polynomial squeeze to match your reference. (In other words, there is good agreement between the two functions.)
(3) If you expect a large polynomial squeeze to be needed, please ensure your initial parameters for the polynomial morph result in good agreement between your reference and objective functions. One way to obtain such parameters is to first apply a --hshift and --stretch morph. Then, use the hshift parameter for a0 and stretch parameter for a1.
```


However, I think we should just allow `squeeze` to be non-monotonic. Like my previous suggestion, sort the grid (and in parallel the function values). Since we have discrete data, overlap is unlikely. In the case of overlap, take the average value of the overlaps. We would no longer have this error. I suggest warning the user if the final (only the final) squeeze parameters results in non-monotonicity and allowing them to disable it with another setting:

```
Warning: The squeeze morph has interpolated your morphed function from a non-monotonically increasing grid. This can result in strange behavior in the regions {list_of_regions}. To disable this setting, please enable --check-increasing.
```

If the final result is still on a monotonic increasing grid, good! That means only the intermediates were non-monotonic. This change only benefits us since if we do not implement this, the user will see an error on all cases where the grid is non-monotonic, and they recover the same behavior using `check-increasing` (or perhaps a better name for this option).

Use something like [this](https://stackoverflow.com/questions/25264798/checking-for-and-indexing-non-unique-duplicate-values-in-a-numpy-array) to find non-unique values efficiently.

Use `np.diff` and check for negatives to quickly find the region where the interpolation has resulted in non-monotonicity.

Note that the grid size can change, but we implemented have ways to handle that in `morphrgrid.py`.

PR #249, opened 2025-08-18
raise `ValueError` if the the number of regression points is less than the number of parameters in `_residual`

ycexiao:

Address the second item in #241.

Please check the modifications.

### DIFF
--- a/tests/test_refine.py
+++ b/tests/test_refine.py
@@ -205,8 +205,8 @@ class TestRefine:
             ValueError,
         ) as error:
             refiner.refine(*refpars)
-            actual_error_message = str(error.value)
-            assert actual_error_message == expected_error_message
+        actual_error_message = str(error.value)
+        assert actual_error_message == expected_error_message
 
         # call from command line
         import subprocess


Sparks29032:
Make sure that the argument parser captures the error in `morphapp.py`. Since the parser works on command line, you want to use its inbuilt error function (though I believe there is also a custom error function). The end result on the CLI side should be to throw `exit 1` rather than a Python error, but the Python side should be fine now.

I would also like it if there was a test that this error is indeed being captured? Maybe one test for the CLI side and one for the morphpy side?

sbillinge:
aren't we raising an exception?  Are saying the exception gets caught and handled somewhere higher up in the code stack?  If not then it will crash with the exception (which would be the desired behavior)?

Sparks29032:
Sounds good for the adjustment. Would also be helpful to mention the size of the function and number of parameters you are optimizing to the user so they can see exactly how many more parameters they must reduce by.

tests/test_refine.py
```diff
@@ -181,6 +181,64 @@ def stretch(x, y, stretch):
...
+            "reduce the number of parameters."
+        )
+        with pytest.raises(
+            ValueError,
+        ) as error:
+            refiner.refine(*refpars)
+            actual_error_message = str(error.value)
+            assert actual_error_message == expected_error_message
+
+        # call from command line
```
ycexiao, line 213: The error is already captured in `morphapp.py` and sent to `custom_error`. So it also works if we don't exit the program in `refine.py`

```
# in `morphapp.py`
try:
    # call refine.py
except ValueError as e:
    parser.custom_error(str(e))
```
An example output:
```
Number of shared grid points: 2
Number of parameters: 3
Not enough shared grid points between the morphed function and the target function to fit the chosen parameters.
Please make sure the overlapping domain between the morphed function and the target function is sufficiently large, or reduce the number of parameters.
```

ycexiao (commented):
@sbillinge @Sparks29032, it's ready for review.

src/diffpy/morph/refine.py
```diff
@@ -85,9 +85,13 @@ def _residual(self, pvals):
         rvec = _y_target - _y_morph
         if len(rvec) < len(pvals):
             raise ValueError(
+                f"\nNumber of shared grid points: {len(rvec)}\n"
+                f"Number of parameters: {len(pvals)}\n"
                 "Not enough shared grid points between morphed function "
-                "and target function to fit the chosen parameters. "
-                "Please adjust the functions or "
+                "between morphed function and target function to fit "
```
sbillinge: please check, is "between morphed function" repeated here?

sbillinge (commented):
nice work everyone

src/diffpy/morph/refine.py
```diff
@@ -83,7 +83,13 @@ def _residual(self, pvals):
             self.x_morph, self.y_morph, self.x_target, self.y_target
         )
         rvec = _y_target - _y_morph
-
+        if len(rvec) < len(pvals):
+            raise ValueError(
+                "Not enough shared grid points between morphed function "
+                "and target function to fit the chosen parameters. "
+                "Please adjust the functions or "
```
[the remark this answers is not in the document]
ycexiao: The current error message is: "Please make sure the overlapping domain between the morphed function and the target function is sufficiently large, or reduce the number of parameters."

src/diffpy/morph/refine.py
ycexiao: Yes, it is a typo. It is fixed in a newer commit.

ycexiao (commented):
@sbillinge @Sparks29032, it's ready for review.

sbillinge (commented):
I am not sure I like this.  Why are we doing this?  Lot's of extra complicated code to maintain, but what benefit does it get us exactly?  Let's discuss a bit.   I am happy to be convinced, but in general, simpler is better for a research group like ours.....

Sparks29032:
Two main reasons why I'd like this PR:
(1) When this error occurs, a quite unhelpful `leastsq` error message pops up and it is not too clear what one needs to fix in the case that the user really has an exorbitant number of parameters say in their `funcxy` function.
(2) When I ran into this error, it was because I was reading the transpose of the two-column array lists. In other words, rather than getting 4000 grid points, I am getting 1. Seeing this error and letting me know that my grid overlap is some absurdly small number like 1 or 0 lets me know that the error has to do with my data preprocessing/obtaining, not some weird regressor fault, which a random `leastsq` error message may imply to a user who is not too familiar w/ debugging.

ycexiao:
> I am not sure I like this. Why are we doing this? Lot's of extra complicated code to maintain, but what benefit does it get us exactly? Let's discuss a bit. I am happy to be convinced, but in general, simpler is better for a research group like ours.....

@sbillinge, are you referring to the commit "[use a custom error to stop leatsq](https://github.com/diffpy/diffpy.morph/pull/249/commits/91fa347f8e783bdb3be5423eabdfd7b234ff63ff)"? Actually, it is not a good fix and it is reverted. The actual modification is from [chore: fix the code indentation.](https://github.com/diffpy/diffpy.morph/pull/249/commits/daf51c3a621156938ea6646d3d1384dbda47a9bf) where only the code indentation is changed so it is not complicated.

Issue #243, opened 2025-07-15
Raise warnings for extreme extrapolation or interpolation


Sparks29032:
- [ ] Give extrapolation warnings based on `low_extrap` and `high_extrap` in `MorphSqueeze.py`
- [ ] Give warnings in `MorphRGrid` when the average grid density is significantly smaller than the maximum grid step (we should have a conversation about the threshold and nature of this warning)

Sparks29032:
Let's close this for now, since the `MorphRGrid` discussion is a bit nuanced. I will open a new issue.

Issue #233, opened 2025-07-09
Clean up shared test functions


Sparks29032:
If any code is reused in tests, put it all in a function. If a particular function is shared across multiple test files, put it in the conftest.

Rethink names of functions (e.g. isfloat -> is_float, are_files_same -> files_same). This is to mirror `numpy`'s `all_close` function naming.

sbillinge:
@Sparks29032 please can you review what is left to do here.  I see some linked merged PRs.  If we summarize that, this will be easier to return to later.  Then we can move the rest to `0.4.0`

Sparks29032:
I don't think anything is actually resolved in this issue.

The linked PRs were motivation for this issue since there was a pretty serious bug with `smear-pdf` that I noticed when playing around.

We will need to take some time for each morph and think of edge cases.

sbillinge:
ok, I moved to `0.4.0`

PR #250, opened 2025-08-25
feat: add warning for extrapolation in `morphsqueeze`

ycexiao:

Address the issue mentioned in #243. Create a warning for extrapolation in `morphsqeeze`.

Please check the implementation.

### DIFF
--- a/src/diffpy/morph/morphs/morphsqueeze.py
+++ b/src/diffpy/morph/morphs/morphsqueeze.py
@@ -4,6 +4,7 @@ function."""
 import numpy as np
 from numpy.polynomial import Polynomial
 from scipy.interpolate import CubicSpline
+import warnings
 
 from diffpy.morph.morphs.morph import LABEL_GR, LABEL_RA, Morph
 
@@ -85,4 +86,17 @@ class MorphSqueeze(Morph):
         high_extrap = np.where(self.x_morph_in > x_squeezed[-1])[0]
         self.extrap_index_low = low_extrap[-1] if low_extrap.size else None
         self.extrap_index_high = high_extrap[0] if high_extrap.size else None
+
+        low_extrap_x = x_squeezed[0] - self.x_morph_in[0]
+        high_extrap_x = self.x_morph_in[-1] - x_squeezed[-1]
+        if low_extrap_x > 0 or high_extrap_x > 0:
+            wmsg = "Extrapolating the morphed function: "
+            if low_extrap_x > 0:
+                wmsg += f"extrapolating length in the lowe r {low_extrap_x} "
+            if high_extrap_x > 0:
+                wmsg += f"extrapolating length in the high r {high_extrap_x} "
+            warnings.warn(
+                wmsg,
+                UserWarning,
+            )
         return self.xyallout


ycexiao:
I will add a test for it if we are sure this implementation is what we want.

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -85,4 +87,17 @@ def morph(self, x_morph, y_morph, x_target, y_target):
         high_extrap = np.where(self.x_morph_in > x_squeezed[-1])[0]
         self.extrap_index_low = low_extrap[-1] if low_extrap.size else None
         self.extrap_index_high = high_extrap[0] if high_extrap.size else None
+
+        low_extrap_x = x_squeezed[0] - self.x_morph_in[0]
+        high_extrap_x = self.x_morph_in[-1] - x_squeezed[-1]
+        if low_extrap_x > 0 or high_extrap_x > 0:
+            wmsg = "Extrapolating the morphed function: "
+            if low_extrap_x > 0:
+                wmsg += f"extrapolating length in the lowe r {low_extrap_x} "
```
Sparks29032, line 96: Typo?

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -85,4 +87,17 @@ def morph(self, x_morph, y_morph, x_target, y_target):
         high_extrap = np.where(self.x_morph_in > x_squeezed[-1])[0]
         self.extrap_index_low = low_extrap[-1] if low_extrap.size else None
         self.extrap_index_high = high_extrap[0] if high_extrap.size else None
+
+        low_extrap_x = x_squeezed[0] - self.x_morph_in[0]
```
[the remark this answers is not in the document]
Sparks29032: I think the best would just be to say all points below `low_thresh` and above `high_thresh` are extrapolated using cubic spline etc. etc.

Best to mention the method of extrapolation used (cubic spline I believe)

src/diffpy/morph/morphs/morphsqueeze.py
sbillinge: @Sparks29032 is the implement ok, typo aside?  Then @ycexiao can write a test.

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -85,4 +94,22 @@ def morph(self, x_morph, y_morph, x_target, y_target):
         high_extrap = np.where(self.x_morph_in > x_squeezed[-1])[0]
```
ycexiao, line 94: Do we want to remove these few lines of code? I can't find other references to the variables defined here.

Sparks29032:
Can you also test it for the CLI case?

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -85,4 +94,22 @@ def morph(self, x_morph, y_morph, x_target, y_target):
...
         self.extrap_index_low = low_extrap[-1] if low_extrap.size else None
         self.extrap_index_high = high_extrap[0] if high_extrap.size else None
+
+        begin_end_sqeeze = min(x_squeezed), max(x_squeezed)
+        begin_end_in = min(self.x_morph_in), max(self.x_morph_in)
+        if not (
+            begin_end_sqeeze[0] <= begin_end_in[0]
+            and begin_end_in[-1] >= begin_end_in[-1]
+        ):
+            wmsg = (
```
ycexiao, line 104: Output effect:
image

ycexiao (commented):
@Sparks29032 @sbillinge, it's ready for review.

src/diffpy/morph/morphs/morphsqueeze.py
Sparks29032: Can you make it say `Warning:` on the CLI as well?

Sparks29032:
I liked the behavior before where you only say extrapolatipn below if there is extrapolation below and above if there is above. So if only the above side has the warning:

Warning: Points with grid values above {begin_end_sqeeze[1]} will be extrapolated via cubic spline.

src/diffpy/morph/morphs/morphsqueeze.py
Sparks29032: These were variables we made before to store what you have now called begin_end_squeeze.

ycexiao:
> We don't need to tell them that we are obtaining points between x and y, as a user of Squeeze should already know this from the docs or manual.

I think it would be helpful to reduce the bandwidth for users when they are tuning the coefficient to choose a moderate extrapolation length if we have both the expected x range and actual x range.

ycexiao:
Thank you! Good to know.

src/diffpy/morph/morphs/morphsqueeze.py
ycexiao: I find these variables provide a handy interface to test the extrapolated effect in `test_morphsqueeze`, so I think we should keep them.

tests/test_morphsqueeze.py
```diff
@@ -85,3 +87,83 @@ def test_morphsqueeze(x_morph, x_target, squeeze_coeffs):
...
+    )
+    run_cmd = ["diffpy.morph"]
+    run_cmd.extend(["--squeeze=" + ",".join(map(str, coeffs))])
+    run_cmd.extend([str(morph_file), str(target_file)])
+    run_cmd.append("-n")
+    result = subprocess.run(run_cmd, capture_output=True, text=True)
+    assert expected_wmsg in result.stderr
+
+
+def create_morph_data_file(
```
ycexiao, line 154: Do we want to create a `helper.py` module in `\tests`? If we want to add more CLI tests,  we might need a handy function to create temporary `*.cgr` files. We can write this function in `helper.py`, and other tests can import it.

ycexiao (commented):
@Sparks29032 @sbillinge, it's ready for review.

ycexiao:
Warning output:
image

tests/test_morphsqueeze.py
Sparks29032: That would be helpful! (see #233)

PR #251, opened 2025-09-02
Remove subprocess calls from tests

Sparks29032:
New tests from #249 and #250 give errors on Windows. Running as shell command (in the case of Windows) is not recommended.

We don't need to call the subprocess since we can directly simulate a CLI call similar to the tests already present in `test_morphapp.py`.

### DIFF
--- a/news/no-subprocess.rst
+++ b/news/no-subprocess.rst
@@ -1,0 +1,23 @@
+**Added:**
+
+* <news item>
+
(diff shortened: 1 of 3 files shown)


sbillinge:
yes, this is much better.  It is also possible to import the `main` function from the entry-point app and run that the desired inputs.  This is how we would normally do it to be platform independent.

PR #252, opened 2025-09-03
Only warn user about `squeeze` extrapolation after refinement

Sparks29032:
Update to #250, which would throw a warning each step of the refinement. We only want the user to see the warning afterward since only that quantity is useful.

Targets first check point of #243


Sparks29032:
Merging and reverting to run full stack of OS tests.

Issue #256, opened 2025-09-13
Allow `squeeze` to search over non-monotonic polynomials


Sparks29032:
I think we should just allow `squeeze` to be non-monotonic. Like my previous suggestion, sort the grid (and in parallel the function values). Since we have discrete data, overlap is unlikely. In the case of overlap, take the average value of the overlaps. We would no longer have this error. I suggest warning the user if the final (only the final) squeeze parameters results in non-monotonicity and allowing them to disable it with another setting:

```
Warning: The squeeze morph has interpolated your morphed function from a non-monotonically increasing grid. This can result in strange behavior in the regions {list_of_regions}. To disable this setting, please enable --check-increasing.
```

If the final result is still on a monotonic increasing grid, good! That means only the intermediates were non-monotonic. This change only benefits us since if we do not implement this, the user will see an error on all cases where the grid is non-monotonic, and they recover the same behavior using `check-increasing` (or perhaps a better name for this option).

Use something like [this](https://stackoverflow.com/questions/25264798/checking-for-and-indexing-non-unique-duplicate-values-in-a-numpy-array) to find non-unique values efficiently.

Use `np.diff` and check for negatives to quickly find the region where the interpolation has resulted in non-monotonicity. Specifically, for each set of continuous negative values in `np.diff(a)` for some array `a`, your non-monotonic region is `a[idx2+1]` to `a[idx1]` where `idx1` is the index of where the negative region in `np.diff(a)` starts and and `idx2` is the index where it ends. For example, if `np.diff(a)` is negative on the range `10-20`, your non-monotonic region is `a[21]` to `a[10]`. You will not have index out of bounds errors from this since `np.diff` returns an array one smaller than the original.

Note that the grid size can change, but we implemented ways to handle that in `morphrgrid.py`.

sbillinge:
I am ok with this.  It is a bit odd, but if there is a unique mapping from domain to the function value, I  guess it is ok, right?

Sparks29032:
Closed by #272

PR #259, opened 2025-09-17
feat: add `--check-increase` option

ycexiao:

Implement some of the functionalities mentioned in #256

`--check-increase` option is added for `squeeze morph`. When it is applied, the non-strictly-increasing x will raise an error; otherwise, it will only throw warnings.

Please check the implementations.

### DIFF
--- a/src/diffpy/morph/morphs/morphsqueeze.py
+++ b/src/diffpy/morph/morphs/morphsqueeze.py
@@ -1,6 +1,7 @@
 """Class MorphSqueeze -- Apply a polynomial to squeeze the morph
 function."""
 
+import numpy
 from numpy.polynomial import Polynomial
 from scipy.interpolate import CubicSpline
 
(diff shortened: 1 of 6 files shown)


ycexiao:
Tests for the expected squeeze morph outcome with non-strictly-increasing `x_squeezed`  are left. Since this needs more experience in identifying what morphed outcome is "good enough", @Sparks29032, can you maybe code this part in another PR?

Sparks29032:
@ycexiao Can you send screenshots of the outputs?

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -68,8 +69,83 @@ class MorphSqueeze(Morph):
...
+    def __init__(self, config=None, check_increase=False):
         super().__init__(config)
+        self.check_increase = check_increase
+
+    def _set_squeeze_info(self, x, x_sorted):
+        self.squeeze_info = {"monotonic": True, "overlapping_regions": None}
+        if list(x) != list(x_sorted):
+            if self.check_increase:
+                raise ValueError(
+                    "Error: The polynomial applied by the squeeze morph has "
```
ycexiao, line 81: The error message is updated according to #248

ycexiao (commented):
@Sparks29032, it's ready for review.

src/diffpy/morph/morphs/morphsqueeze.py
```diff
@@ -68,8 +69,68 @@ class MorphSqueeze(Morph):
...
+        overlapping_regions_sign = -1 if x[0] < x[-1] else 1
+        overlapping_regions_x = [
+            monotomic_regions[i]
+            for i in range(len(monotomic_regions))
+            if monotomic_signs[i] == overlapping_regions_sign
+        ]
+        overlapping_regions = [
+            (min(region), max(region)) for region in overlapping_regions_x
+        ]
+        return overlapping_regions
```
Sparks29032: This is not efficient. We only need the ends of the non-monotonic regions, not all the grid points of the region.

See my comment in #248 for implementation. We do not want to iterate over the entire list.

Here is a suggested approach using only numpy (to ensure all computations are done at the C-level rather than Python).
```
# Given a grid x
diffx = np.diff(x)
signx = np.sign(diffx)

# Create array that is 1 when signx is -1 and 0 otherwise. Padding added.
signx_negative = np.concatenate(([0], (signx==-1).astype(int), [0]))

# Left sides of our intervals
indices_left = np.where(np.diff(signx_negative)==-1)[0]

# Right sides of our intervals
indices_right = np.where(np.diff(signx_negative)==1)[0]

# Get x grid values corresponding to the interval indices
x_vals_left = x[indices_left]
x_vals_right = x[indices_right]

# Get the intervals
overlap_intervals = list(zip(x_vals_left, x_vals_right))
```

Sparks29032:
@ycexiao See my in-line comment about implementing the overlap region search. Also, I would prefer if we had tests for your three helper functions.

ycexiao:
> When the --check-increase function is enabled, it should fail. Take off the --check-increase function and see if we converge to the same value.

The turn-on and off behavior has already been tested by the existing tests. When `--check-increase` is provided, it will throw an error. When it is not by default, none of the existing tests went wrong. 

The tests I left here are whether or not the numerical change, like the change of derivatives,  brought by sorting the non-strictly-increasing x, is acceptable. We don't have a non-strictly-increasing test case before, since it will just raise an error. So I don't have a sense of "ground truth" to compare the current outcome with. But all the existing tests passed, maybe it means that the change is acceptable because the program is able to refine the results using the sorted `x_squeeze`?

ycexiao:
@Sparks29032 @sbillinge, the current status of this PR:

This PR lacks convergence tests for non-strictly-increasing-x. I got stuck on designing the test case for it.

We need a carefully designed test case. The coefficients should be small enough to let all zero guess converge to the non-stricly-increasing coefficients and large enough to avoid extrapolation, as we don't have a convenient way to get extrapolation boundaries from `morph_arrays`, and errors in the extrapolation part are huge.

The extrapolation information is only contained in the printed warning messages. If we can let `morph_arrays` return a Python object containing this information, we can implement this more easily.

sbillinge:
@ycexiao @Sparks29032 can this be closed or backburnered (per the discussion)?

ycexiao:
I think it can be closed. We can implement this feature when the API becomes more flexible or our users ask us to do so.

Sparks29032:
@sbillinge @ycexiao Sorry I missed this thread. After thinking a bit more, I feel the default (safe) feature should be to throw the error (with our custom error message, which is not yet implemented, is stuck in limbo in this pr). Then the user can request to ignore the increasing with an option (with our warning). Since we haven't fleshed the math out for convergence / error bounds, this may be the safest approach?

sbillinge:
@Sparks29032 I am not sure the best approach.  IT really depends how common this is and how harmful.  When you make software harder to use it can frustrate people so if someone is just trying  to morph their data to look at something at the beamline and it keeps not working that is bad.  If we are saving them from something bad, that is fine, but if it is some trivital thing that just doesn't really matter, that is less good.

1. Is this something that only happens once every 50 years, or is it happening like 30% of the time?
2. If we shuffle the points and don't blow up the run, I still don't have a very good feeling about what is happening then, that seems a bit odd, is it modifying the data in some rather bad and irreversible way or is it just kind of aesthetics?

Of course merphing itself is modifying data, so it is kind of "photoshop" for data so users are already venturing into dangerous territory.

Do we have thoughts about 1. and 2.?

sbillinge:
closing as per @Sparks29032 comment above.  Replaced by #272

PR #274, opened 2025-12-17
Handle IO for (CLI) morph inputs in a function

Sparks29032:
CLI morph inputs were handled separately in the single morph and two multiple morph functions within `morphapp.py`. This has been restructured into a function in `morph_io.py`. Note that Python-only morphs are still handled separately outside this function as they are not enabled for `multiple-morphs/targets` (since a user can simply write a for loop instead).


sbillinge (commented):
this looks fine modulo comments.  It needs tests for the logic.

PR #299, opened 2026-04-24
Add uncertainties for main parameters.

### LINKED ISSUE #300: Add option to see uncertainties
### Problem

Give user the option `-u` to display uncertainties for fitted morphing parameters.

### Proposed solution

See #299

Sparks29032:
Closes #300 

Need tests for warnings and each parameter separately (including baselineslope).

Need special handling for `squeeze` and `func[x/y/xy]` morph parameters.

image


Sparks29032:
Case where all parameters have uncertainties:
image

Case where certain parameters cause the Hessian to turn singular:
image

Tests and documentation updated. Ready to merge after review.

Sparks29032:
Oops, previous one I ran only one of the tests I thought would be affected... fixed now.

Sparks29032:
image

Loosened tolerance for checking numerics match. New algorithms and updates to old ones can cause some deviation (e.g. works on one of our tests on PR workflows but not others).

src/diffpy/morph/morph_io.py
```diff
@@ -399,11 +422,16 @@ def multiple_morph_output(
                 output = f"\n# Target: {target}\n"
             else:
                 output = f"\n# Morph: {target}\n"
-            output += "# Optimized morphing parameters:\n"
-            output += "\n".join(
-                f"# {param} = {morph_results[target][param]:.6f}"
-                for param in morph_results[target]
-            )
+            # output += "# Optimized morphing parameters:\n"
```
sbillinge: are you retaining these lines for some reason?  If yes, state the reason and when they can be removed, and if not, then just remove them.

sbillinge (commented):
looks great.  One inline comment.

src/diffpy/morph/morph_io.py
Sparks29032: Ah, thanks. I left that there as reference for a different approach originally.

Sparks29032:
@sbillinge Removed the stray comment.

Since pre-commit isn't working on here (we will address by #301 hopefully), this screenshot is just to confirm it is passing it locally.
image

sbillinge:
@Sparks29032 Please could you fix pre-commit.  Check `scikit-package` for the update.  We need to update versions on pretty much everything, but there are specific changes to docformatter that should fix this bug and get tests passing.

sbillinge:
this still seems to be  failing tests @Sparks29032 .  I didn't look why.

Sparks29032:
Force-push to remove rebase. Something went wrong there.

Sparks29032:
No more conflicts now. Good to merge.

PR #304, opened 2026-06-01
Give user option to save on original grid

### LINKED ISSUE #284: Let user choose which grid the morph is returned on
If a user morphs `objective.gr` to match `target.gr`, the returned morphed function `morph.gr` will be on the grid of `target.gr`. The user should be able to select whether they want this or for `morph.gr` to be returned on the original grid of `objective.gr`. This can be done by creating a parameter to toggle the last application of `MorphRGrid()`.

Sparks29032:
Closes #284.

Default right now is to save the morphed function on the intersected grid. This is the same grid as `--diff` returns the function on as the subtraction occurs on the intersected grid. This PR allows the user to select `--original-grid` to return the morphed function on the original grid the morph happened on.

Tests are written for both the CLI and MorphPy.


Sparks29032:
Good to merge after review. Pretty much only one line has changed.

image

Issue #211, opened 2025-06-16
Create a general smearing morph


Sparks29032:
### Problem

Generalize the smear morph to not just Gaussians.

Sparks29032:
I am requesting this for comparing DOS curves at different temperatures (Lorentzian broadening).

sbillinge:
I moved it back to 0.5.0 release.  Do you want to do it @Sparks29032 or shall we give it to one of the new people?

Sparks29032:
@sbillinge I'm writing the one for Lorentzian since I would like to use it. The second part (letting the user pass an arbitrary function to be smeared) I will probably not work on, so a new person can feel free to take a crack at it.

Edit: I put some notes on how this arbitrary function should be implemented.

PR #309, opened 2026-07-09
Add `--smear-function` option and Lorentzian smearing

Sparks29032:
Addresses part of #211. This adds a Lorentzian smearing option.

We will need another PR that allows the user to pass a general smearing function, but the outline for how to do that is already done in this PR and should be relatively straightforward to implement from here.

The default if `--smear-function` is not enabled is still Gaussian, no no workflow changes will have to be made for those using the Gaussian smearing.

Below are two morphs. The morphed function is a Lorentzian with HWHM `0.1`. The target is a Lorentzian with HWHM `0.25`. The first image is when Gaussian smearing (and scaling) is applied. The second is when Lorentzian smearing (and scaling) is applied.

Gaussian smearing:
image

Lorentzian smearing:
image


tests/test_morphio.py
```diff
@@ -466,3 +468,102 @@ def fx(x, y, ax0, ax1):
         for unc in morph_info["uncertainties"].keys():
             assert unc in params
             assert morph_info["uncertainties"][unc] is not None
+
+    def test_smear_func(self, setup, tmp_path):
+        file = self.testfiles[-1]
+        sfn = "smear_morph_result.cgr"
+        save_file = (tmp_path / sfn).resolve().as_posix()
+
+        # Test Gaussian default
```
sbillinge: per group standard, please specify what the expected behavior is for each case, something like
`# Test Gaussian default.  Expect Gaussian convolution is correctly applied`

tests/test_morphio.py
```diff
@@ -466,3 +468,102 @@ def fx(x, y, ax0, ax1):
...
+        )
+        pargs = [file, file]
+        single_morph(self.parser, opts, pargs, stdout_flag=False)
+        with open(test_saving_smear.joinpath("gaussian.cgr")) as tf:
+            with open(tmp_path.joinpath(sfn)) as gf:
+                actual = filter(ignore_path, gf)
+                expected = filter(ignore_path, tf)
+                are_files_same(actual, expected)
+
+        # Test explicit Gaussian specification with capitalization
```
sbillinge: see above and for all the cases

sbillinge (commented):
one small gripe, but a great PR, thanks!

Sparks29032:
@sbillinge Updated headers. Tests for the i/o are in `test_morphio.py` and only check if the file headers produced are correct and that the new option is handled correctly. No morphs are applied (`apply=True` with `smear=0` in those tests). The actual test for the Lorentzian morph behavior is in `test_morphsmear.py`.

PR #311, opened 2026-07-28
skpkg: recut morph package

stevenhua0320:
@sbillinge ready to review. In this PR:
we do a recut for the `diffpy.morph` package. Since it has been `python 3.14`, we will only do the `package update` here.


sbillinge:
This is almost perfect.  Please just update my  email everywhere to `sbillinge@ucsb.edu`.  Then I can merge.

